Issue #4, opened 2015-03-07
add --update option for ament_copyright


tfoote:
Since the tool knows what should be there, it should be able to assist injecting the correct headers.

From: https://github.com/ament/ament_lint/pull/3#issuecomment-77659212

Make sure to use the canonical template from the discussion in #3

dirk-thomas:
The ament_copyright package has been updated with the two referenced PRs.

Regarding the referenced "template": it also includes license information. I think we need a separate package dealing with that.

Issue #7, opened 2015-03-09
improve Jenkins test result grouping


dirk-thomas:
Choose better testcase classname.

dirk-thomas:
Since I had to test this with the CI job I didn't go through a PR: f1e14e1f38ee4579818099b353f5e1b991a50958

Output before: http://54.183.26.131:8080/job/ros2_batch_ci/94/testReport/

Output now: http://54.183.26.131:8080/job/ros2_batch_ci/95/testReport/

@esteve @tfoote @wjwwood FYI

wjwwood:
Cool, +1

PR #8, opened 2015-03-10
add more options to ament_copyright

dirk-thomas:
The new options are:

* `--list-names` outputs the known copyright holder names and the long string which is injected into files

* `--add-missing "Custom name"` or `--add-missing osrf` which will add the copyright line to all files (Python, C/C++, CMake) which don't have one

* `--add-copyright-year` adds the current year to all files which already have a copyright line

@esteve @tfoote @wjwwood Please review.

Connects to ament/ament_lint#4


dirk-thomas:
Connects to ament/ament_lint#4

ament_copyright/setup.py
```diff
@@ -0,0 +1,36 @@
...
+    classifiers=[
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='Ament is a build system for federated packages.',
+    long_description='''\
+Ament defines metainformation for packages, their dependencies,
+and provides tooling to build these federated packages together.''',
```
esteve: Shouldn't this describe `ament_copyright` and not `ament` in general?

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+        for key in sorted(names.keys()):
+            print('%s: %s' % (key, names[key]))
+        return 0
+
+    if args.xunit_file:
+        start_time = time.time()
+
+    filenames = get_files(args.paths, extensions)
+    if not filenames:
+        print('No files found', file=sys.stderr)
```
wjwwood: Should this be prefixed by some context, e.g. `ament_copyright: No files found`? It might be confusion if you imagine this is the only output mixed in with other output from the other linters or from CMake.

wjwwood: Well there are other places below where I could same something similar, so maybe it's best to leave it be and refine the output where needed.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+            content = content[:end_index] + '-' + current_year + \
+                content[end_index:]
+
+        # output beginning of file for debugging
+        # index = end_index
+        # for _ in range(3):
+        #     index = get_index_of_next_line(content, index)
+        # print('<<<')
+        # print(content[:index - 1])
+        # print('>>>')
```
wjwwood: Should this be ditched or is it important for future iterations on this tool?

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+        inserted_block = '\n' + inserted_block
+    content = content[:begin_index] + inserted_block + content[end_index:]
+
+    # output beginning of file for debugging
+    # index = end_index + len(inserted_block)
+    # for _ in range(3):
+    #     index = get_index_of_next_line(content, index)
+    # print('<<<')
+    # print(content[:index - 1])
+    # print('>>>')
```
wjwwood: Same thing here.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+            continue
+
+        with open(filename, 'r') as h:
+            content = h.read()
+
+        index = scan_past_coding_and_shebang_lines(content)
+        index = scan_past_empty_lines(content, index)
+
+        assert is_copyright_line(content, index), \
+            "Could not find copyright line in file '%s'" % filename
```
wjwwood: Would it be more appropriate to have a `sys.exit` here? Or would we prefer a traceback?

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+
+def scan_past_empty_lines(content, index):
+    while is_empty_line(content, index):
+        index = get_index_of_next_line(content, index)
+    return index
+
+
+def get_index_of_next_line(content, index):
+    index_n = content.find('\n', index)
+    index_r = content.find('\r', index)
```
wjwwood: When processing files from the OS, maybe using `os.linesep` would be possible: https://docs.python.org/2/library/os.html#os.linesep

It's probably ok to not do that for now, but we maybe end up having to patch this on Windows.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+        return len(content)
+    if index_n == -1:
+        return index_r + 1
+    if index_r == -1:
+        return index_n + 1
+    return min(index_n + 1, index_r + 1)
+
+
+def is_comment_line(content, index):
+    return content[index] == '#' or content[index:index + 1] == '//'
```
wjwwood: Does this need to include ` *` for cases like this:

```
/*
 * Copyright ....
 */
```

We can also just agree to not use that style comments for the license and copyright block, but we state the assumption, maybe in a comment or a TODO.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+
+def is_empty_line(content, index):
+    return get_index_of_next_line(content, index) == index + 1
+
+
+def get_comment(filename, msg):
+    if filename.endswith('.py'):
+        line_prefix = '# '
+    else:
+        line_prefix = '// '
```
wjwwood: Same comment about block comments in C.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,31 +12,354 @@
...
+def is_copyright_line(content, index):
+    if not is_comment_line(content, index):
+        return False
+
+    end_index = get_index_of_next_line(content, index)
+    line = content[index:end_index].lower()
+
+    if line.startswith('#'):
+        line = line[1:]
+    elif line.startswith('//'):
```
wjwwood: Same thing. At the very least it might be possible to consolidate this comment related logic.

ament_copyright/ament_copyright/__init__.py
dirk-thomas: One problem is that the tool itself does not know about packages etc. It simply crawls the filesystem. Let's see how the unit test results actually appear in Jenkins (after fixing things like #7).

ament_copyright/ament_copyright/__init__.py
dirk-thomas: I would keep it for now since it is really helpful for debugging.

ament_copyright/package.xml
```diff
@@ -5,15 +5,12 @@
...
   <license>Apache License 2.0</license>
-  <license>BSD</license>
 
-  <buildtool_depend>ament_cmake_core</buildtool_depend>
-  <buildtool_depend>ament_cmake_python</buildtool_depend>
-  <buildtool_depend>ament_cmake_test</buildtool_depend>
+  <buildtool_depend>python3</buildtool_depend>
 
-  <buildtool_export_depend>ament_cmake_test</buildtool_export_depend>
+  <buildtool_export_depend>python3</buildtool_export_depend>
```
wjwwood: Does it make sense to have `python3` be a buildtool dependency? If the intention is to be able to run this python script after building you'll need at least a `run_depend` on a target platform specific version of python.

ament_copyright/setup.py
```diff
@@ -0,0 +1,36 @@
...
+setup(
+    name='ament_copyright',
+    version='0.0.0',
+    packages=find_packages(exclude=['test']),
+    install_requires=['setuptools'],
+    author='Dirk Thomas',
+    author_email='dthomas@osrfoundation.org',
+    maintainer='Dirk Thomas',
+    maintainer_email='dthomas@osrfoundation.org',
+    url='https://github.com/ament/ament_lint/wiki',
```
wjwwood: Are you going to create a wiki for it? Or should you just point to the repo landing page and make use of the `README`?

ament_copyright/ament_copyright/__init__.py
dirk-thomas: This is a good question we should probably answer in a generic way. I usually use `assert` when it is unexpected to happen. Showing the stacktrace to the use should be fine in that case since it should "never" be the case. But thats just my 2 cent.

ament_copyright/setup.py
wjwwood: This looks like a copy-paste which was not updated.

ament_copyright/ament_copyright/__init__.py
wjwwood: That's fine as long as you do not expect it to happen, I interpreted this as something that could happen in general use.

ament_copyright/ament_copyright/__init__.py
dirk-thomas: The problem is that the file can have any of these characters which does not need to correspond with the OS this script is running on.

I can also not use `splitlines()` or something similar here since I need the absolute index in the file.

wjwwood: Yeah, it just seems fragile to things like this:

```
line 1\n\r
line\re 2\n\r
line 3\n
line 4\r\n
```

I'm not saying this will happen, but I was just trying to think adversarially. It's fine by me to leave it for now.

ament_copyright/ament_copyright/__init__.py
dirk-thomas: In order to handle block comments the whole logic would need to be modified (and will become more complex since `is_comment_line()` requires state or must operate on multiple lines). Therefore I would only aim to support `//` as well as `#` for the copyright message.

This should be described in the docs (part of ros2/ros2#3).

ament_copyright/setup.py
dirk-thomas: Indeed: updated in 972244f3c73b5ca207b1bae1a0bbceee9fe0f4fe

ament_copyright/ament_copyright/__init__.py
dirk-thomas: I guess it should check all these combinations then.

ament_copyright/setup.py
dirk-thomas: Updated URL to point to repo only: c3ef45f40b6d46e91c045338c21c97653ea5bc79

ament_copyright/package.xml
dirk-thomas: Removed: 1f62b4765e59e0ff1f3d83c9461d22ecfab3a46e

The only package which still has that is `ament_package`.

wjwwood:
Looks good to me, +1

esteve:
+1

ament_copyright/ament_copyright/__init__.py
dirk-thomas: I added handling \r\n in 8aad6c81caacb4975d3d066a2850e4afa7f7909f

wjwwood:
+1

tfoote:
lgtm

PR #16, opened 2015-04-02
Some fixes to ament_lint_cmake

wjwwood:
Two main things:

- added a `--filters` option to `ament_lint_cmake`
- added the ability to set filters for a file using a CMake comment

@dirk-thomas @esteve @tfoote for review.

### DIFF
--- a/ament_lint_cmake/ament_lint_cmake/cmakelint.py
+++ b/ament_lint_cmake/ament_lint_cmake/cmakelint.py
@@ -16,6 +16,7 @@ https://github.com/richq/cmake-lint/blob/7b85fe46b9bd66fe11ecfef20060f976a49d966
 - allow closing parenthesis to be on the same level as the opening one (lines 281-282)
 - ignore lines with exceeding length if they only contain a single string (lines 35, 196-207)
 - improve SetFilters ability to parse new filters (lines 89-92)
+- implement in file pragmas to control filters on a per file basis (lines 389 - 430)
 """
 from __future__ import print_function
 import sys
--- a/ament_lint_cmake/ament_lint_cmake/cmakelint.py
+++ b/ament_lint_cmake/ament_lint_cmake/cmakelint.py
@@ -15,6 +15,7 @@ https://github.com/richq/cmake-lint/blob/7b85fe46b9bd66fe11ecfef20060f976a49d966
 - removed __version__ import
 - allow closing parenthesis to be on the same level as the opening one (lines 281-282)
 - ignore lines with exceeding length if they only contain a single string (lines 35, 196-207)
+- improve SetFilters ability to parse new filters (lines 89-92)
 """
 from __future__ import print_function
 import sys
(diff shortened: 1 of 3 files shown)


ament_lint_cmake/ament_lint_cmake/cmakelint.py
```diff
@@ -382,6 +386,16 @@ def IsValidFile(filename):
     return filename.endswith('.cmake') or os.path.basename(filename).lower() == 'cmakelists.txt'
 
 def ProcessFile(filename):
+    # Store and then restore the filters to prevent pragmas in the file from persisting.
+    original_filters = list(_lint_state.filters)
+    try:
+        return _ProcessFile(filename)
```
tfoote: Wouldn't this be cleaner to pass in a copy of the filters instead of copying mutating and restoring?

tfoote:
lgtm with the one comment

ament_lint_cmake/bin/ament_lint_cmake
```diff
@@ -109,8 +115,8 @@ def get_files(paths):
 
                 # select files by name  / extension
                 for filename in sorted(filenames):
-                    if filename.lower() == 'CMakeLists.txt'.lower() or \
-                            filename.endswith('.cmake'):
+                    fname_low = filename.lower()
+                    if fname_low == 'CMakeLists.txt'.lower() or fname_low.endswith('.cmake'):
```
dirk-thomas: When would the extension ever not be lowercase?

wjwwood: I guess in the scenario that `CMakeLists.txt` is not capitalized correctly.

dirk-thomas:
LGTM.

Can you create a PR / separate PRs to the upstream repo as well? May be they will be accepted (with changes) then we can keep our copy in sync with upstream.

ament_lint_cmake/ament_lint_cmake/cmakelint.py
wjwwood: I don't think so, because the `_lint_state` object is a global and `_ProcessFile` (potentially) mutates it. If I were to implement this in the `_ProcessFile` function then I would just wrap the whole thing in a try/finally (which is what I've done). Is there something you were thinking I should do instead?

tfoote: Ok. I didn't know that _lint_state might change besides the filter set as that's the only change directly operated upon.

wjwwood:
I'll make the pull requests upstream.

@esteve can you review this?

esteve:
+1

wjwwood:
Here's the first pull request which improves the filters: https://github.com/richq/cmake-lint/pull/5

wjwwood:
Here's the second which allows for in file filtering with a CMake comment: https://github.com/richq/cmake-lint/pull/6

PR #17, opened 2015-04-06
find cppcheck on windows

dirk-thomas:
@esteve @tfoote @wjwwood Please review.

### DIFF
--- a/ament_cppcheck/bin/ament_cppcheck
+++ b/ament_cppcheck/bin/ament_cppcheck
@@ -5,6 +5,7 @@ from __future__ import print_function
 import argparse
 import multiprocessing
 import os
+from shutil import which
 import subprocess
 import sys
 import time
@@ -41,7 +42,12 @@ def main(argv=sys.argv[1:]):
         print('No files found', file=sys.stderr)
         return 1
 
-    cppcheck_bin = find_executable('cppcheck')
+    additional_paths = None
+    if os.name == 'nt':
+        # search in location where cppcheck is installed via chocolatey
+        program_files_32 = os.environ.get('ProgramFiles(x86)', 'C:\\Program Files (x86)')
+        additional_paths = [os.path.join(program_files_32, 'Cppcheck')]
+    cppcheck_bin = find_executable('cppcheck', additional_paths=additional_paths)
     if not cppcheck_bin:
         print("Could not find 'cppcheck' executable", file=sys.stderr)
         return 1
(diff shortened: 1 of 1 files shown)


tfoote:
lgtm

ament_cppcheck/bin/ament_cppcheck
```diff
@@ -41,7 +42,11 @@ def main(argv=sys.argv[1:]):
         print('No files found', file=sys.stderr)
         return 1
 
-    cppcheck_bin = find_executable('cppcheck')
+    additional_paths = None
+    if os.name == 'nt':
+        # search in location where cppcheck is installed via chocolatey
+        additional_paths = ['C:/Program Files (x86)/Cppcheck']
```
esteve: Please use `%ProgramFiles%` instead. `Program Files` is mapped to a localized path in non-English installations of Windows:

http://en.wikipedia.org/wiki/Program_Files

wjwwood:
+1

ament_cppcheck/bin/ament_cppcheck
wjwwood: It would be `%ProgramFiles(x86)%` or both.

wjwwood: The problem is that I don't think that will get expanded by the `which` command, instead you'll need to get `program_file = os.environ.get('ProgramFiles(x86)', "C:/Program Files (x86)/")` and use that in the extra path.

dirk-thomas: Does the environment variable contain a trailing slash?

wjwwood: Doesn't look like it:

screen shot 2015-04-06 at 3 18 22 pm

ament_cppcheck/bin/ament_cppcheck
```diff
@@ -41,7 +42,12 @@ def main(argv=sys.argv[1:]):
         print('No files found', file=sys.stderr)
         return 1
 
-    cppcheck_bin = find_executable('cppcheck')
+    additional_paths = None
+    if os.name == 'nt':
+        # search in location where cppcheck is installed via chocolatey
+        program_files_32 = os.environ.get('ProgramFiles(x86)', 'C:/Program Files (x86)')
```
wjwwood: `\` rather than `/`

ament_cppcheck/bin/ament_cppcheck
```diff
@@ -41,7 +42,12 @@ def main(argv=sys.argv[1:]):
...
+        additional_paths = [os.path.joni(program_files_32, 'Cppcheck')]
```
wjwwood: `join` not `joni`.

wjwwood:
Other than the comments +1

esteve:
+1

dirk-thomas:
Finally passed on the farm: http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_windows_cppcheck/8/

PR #21, opened 2015-04-13
add env var to prevent ament_cppcheck from using -j#

wjwwood:
For some unknown reason, if `-j` is passed to `cppcheck` it immediately exits docker on exit (both when successful and when failing), stopping the CI build prematurely.

I've added a check for an environment variable, `AMENT_CPPCHECK_NO_MULTI_THREADED `, which when set will prevent `ament_cppcheck` from calling `cppcheck` with `-j`, working around the issue until we can figure out why.

I would do something more fully fledged, like passing this setting in from outside using a CMake variable, but I expect that we will replace this workaround with an actual fix or some other heuristic when we learn more.

I will put together a minimal Dockerfile and report this issue to upstream.

@dirk-thomas @esteve @tfoote for review

### DIFF
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -82,7 +82,7 @@ def main(argv=sys.argv[1:]):
            '-rp',
            '--xml',
            '--xml-version=1']
-    if jobs:
+    if jobs and 'AMENT_CPPCHECK_NO_MULTI_THREADED' not in os.environ:
         cmd.extend(['-j', '%d' % jobs])
     cmd.extend(files)
     try:


wjwwood:
Testing this branch here: http://54.183.26.131:8080/job/ros2_batch_ci_test_linux/6/console

wjwwood:
Here is an example of it stopping dead in it's track on the `cppcheck` for `rclcpp`: http://54.183.26.131:8080/job/ros2_batch_ci_test_linux/5/console

Both @tfoote and I can reproduce this result locally.

tfoote:
+1

wjwwood:
I had to restart the job, there was bug with how I passed the options down to the job through docker:

http://54.183.26.131:8080/job/ros2_batch_ci_test_linux/7/console

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -82,7 +82,7 @@ def main(argv=sys.argv[1:]):
            '-rp',
            '--xml',
            '--xml-version=1']
-    if jobs:
+    if jobs and 'AMENT_CPPCHECK_NO_MULTI_THREADED' not in os.environ:
```
dirk-thomas: Please add a comment above this line referencing this ticket.

esteve:
+1

wjwwood:
Ok, I found the root cause of the issue. The way I was invoking the docker CMD prevented the job server from working. I will just close this, as it is not needed anymore.

PR #25, opened 2015-08-21
disable multi-line comment formatting since the result is just not good

dirk-thomas:
See ros2/rmw_connext#94 how the comments needed to look before to be compliant with uncrustify.


tfoote:
With this disabled you can add the stars in front of each line?

jacquelinekay:
you can also make sweet ASCII art comments without uncrustify ruining the identation

```
/*
             ____ 
    .__     /_   |
  __|  |___  |   |
 /__    __/  |   |
    |__|     |___|
*/
```

esteve:
+1

@jacquelinekay :laughing:

dirk-thomas:
@tfoote Why would prepend every line with a star? But yes, you can do that now.

tfoote:
I was wondering what this changed in the enforcement. A quick search didn't find a definition of the flag. I was referring to enclosures such as [here](http://syque.com/cstyle/ch4.6.htm)

PR #29, opened 2015-09-17
update documentation for linters

dirk-thomas:
Based on feedback in https://github.com/ament/ament_lint/pull/28#discussion_r39809234

Connect to ament/ament_cmake#35


wjwwood:
lgtm

PR #31, opened 2015-10-14
update cpplint configuration

dirk-thomas:
The patch also implements a custom include guard pattern which follows the Google style guide with only one difference: instead of a single separator between folder names / namespaces and filenames it uses two. This ensures that separate packages have unique include guards.

The following example would otherwise in the same colliding include guards:
* package `foo_bar` has a file `baz.hpp`
* package `foo` has a file `bar_baz.hpp`

ament/ament_index#6 is an example what we need to update to make our code compliant with cpplint.

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -28,7 +28,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# https://code.google.com/p/google-styleguide/source/browse/trunk/cpplint/cpplint.py?r=141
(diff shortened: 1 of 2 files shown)


wjwwood:
lgtm

dirk-thomas:
@esteve @jacquelinekay @tfoote Do you all agree to follow this style?

If yes, I will update the style guide in the wiki accordingly and we can start updating existing code to pass this linter before enabling it globally.

jacquelinekay:
Alright, I pledge to follow the Google C++ style guidelines again.

Migrating the codebase back to this style will be an interesting process. Do we plan on doing it gradually or in one pass?

wjwwood:
It shouldn't be too different from our current style. These addendums to the google style guide should be mentioned in our development guide: https://github.com/ros2/ros2/wiki/Developer-Guide#c-1 (which I believe Brain has gone through and made up to date recently, but if it needs more changes then they should be made since it's linked from our main ros2 wiki).

tfoote:
+1 to doing a double underscore between package name and file path. 

It would be good to do a focused effort to get it migrated so we can turn on the automated checking soon.

PR #33, opened 2015-10-25
ignore cpplint runtime/int error

dirk-thomas:


dirk-thomas:
Please review.

esteve:
Wouldn't it be better to actually change `int` / `long` to sized integers in our code? (e.g. int32, int64, etc.)

dirk-thomas:
That would be an alternative. Could you work on create PRs for that instead? Then we don't have to ignore this cpplint warning.

esteve:
Is there a log of cpplint on the buildfarm to see the warnings?

dirk-thomas:
No, I only ran the tests with ament_cmake_cpplint added as a default linter locally.

tfoote:
We'll merge this as is. And open a ticket to clean up the usage and let us revert this later.

Issue #34, opened 2015-10-27
reenable runtime/int for cpplint


dirk-thomas:
Related to #33.

Existing code need to be updated to pass.

esteve:
CI Linux job:

http://ci.ros2.org/job/ros2_batch_ci_linux/492/

dirk-thomas:
The CI job has no effect since `ament_cpplint` is not enabled as a common linter.

esteve:
But it failed, so running the CI job is always a good a idea. Anyway, I'll fix the compile errors.

dirk-thomas:
I expected you to compile the code locally before creating the PRs. In order to run `ament_cpplint` in the CI it needs to be added to https://github.com/ament/ament_lint/blob/master/ament_lint_common/package.xml But several packages will have other cpplint errors until the pending items on https://github.com/ros2/ros2/issues/12 are addressed.

esteve:
I expected it wouldn't fail, but mistakes happen.

esteve:
Errors fixed and compiles locally.

PR #35, opened 2015-10-28
Reenable int/long check

esteve:
Reenable int/long check

Connects to #34


dirk-thomas:
Can you please link the PRs resolving this for all the repos?

esteve:
@dirk-thomas they are linked in #34

dirk-thomas:
Then we should move this ticket back to in-progress since i don't think `system_tests` is the only repo which needs changes.

PR #37, opened 2015-11-05
enforce single line comments for closing namespaces

dirk-thomas:
Also fix an infinite loop when running on files not in a vcs working copy.


esteve:
+1

PR #38, opened 2015-11-05
fix issue with latest pyflakes release

dirk-thomas:


tfoote:
+1, for the record can you link to the cause of the error, or manifestation?

dirk-thomas:
This was not ready to be reviewed but supposed to fix the test failures of ament/ament_tools#65.

PR #39, opened 2015-11-07
use Python 3 for cpplint and split into Python and CMake package

dirk-thomas:
This applies the pending PR from cpplint for Python 3 compatibility: google/styleguide#47

Fixes ament_cpplint on OS X and Windows:
* http://ci.ros2.org/job/ros2_batch_ci_linux/576/
* http://ci.ros2.org/job/ros2_batch_ci_osx/458/
* http://ci.ros2.org/job/ros2_batch_ci_windows/615/


ament_cpplint/test/test_pep257.py
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ament_pep257.main import main
+
+
+def test_pep257():
+    excluded = ['ament_cpplint/cpplint.py']
```
jacquelinekay: Just wondering, why does this test exclude `ament_cpplint/cpplint.py` while the others just exclude `cpplint.py`? Are they all supposed to be the same path?

dirk-thomas: Different tools, different API :wink:

jacquelinekay:
+1

PR #41, opened 2015-11-14
avoid using message without STATUS

dirk-thomas:
Because `message("...")` goes to `stderr`.

Connect to ament/ament_cmake#45


tfoote:
+1

esteve:
+1

PR #42, opened 2015-11-22
remove result type extension from testsuite name

dirk-thomas:
Since #27 added the result type extension to make Jenkins happy all the testsuite names had the extra extension in their name. This added an unnecessary level to the hierarchy.

Before: http://ci.ros2.org/job/ros2_batch_ci_linux/621/testReport/ament_clang_format.lint_cmake/
After: http://ci.ros2.org/job/ros2_batch_ci_linux/622/testReport/ament_clang_format/lint_cmake/

### DIFF
--- a/ament_clang_format/bin/ament_clang_format
+++ b/ament_clang_format/bin/ament_clang_format
@@ -185,6 +185,9 @@ def main(argv=sys.argv[1:]):
         suffix = '.xml'
         if file_name.endswith(suffix):
             file_name = file_name[0:-len(suffix)]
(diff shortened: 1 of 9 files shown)


dirk-thomas:
Please review.

tfoote:
+1, it would be nice to reuse this snippet since it's identical in 9 places including several lines before ad after the changes, but they don't have a common dependency related to linting at the moment so it's not worth adding a dependency. But if there's more in the future it might be worth adding somewhere to contain similar common methods. .

PR #43, opened 2015-12-11
ignore space after a semi colon

wjwwood:
Sometimes it makes sense to not have whitespace after a semicolon, e.g. `if (true) {return foo;}`. Other times it is appropriate, e.g. `#define FOO(arg) bar(); foo(arg); baz();`. Since `uncrustify` cannot know when it is appropriate I propose we tell it to ignore this rule and we can leave this to the developer's discretion. Since it is ignored, existing code should not be affected, but I could go through and find places it really should not have space after it, but I don't think it matters that much.


dirk-thomas:
lgtm

PR #44, opened 2016-02-09
update config to uncrustify 0.62

dirk-thomas:
Connect to ament/uncrustify#12


wjwwood:
This is basically a merge with the new upstream default?

lgtm

dirk-thomas:
Yes, the config file was updated with the new options and some values have been selected to be similar to our other values.

PR #45, opened 2016-02-11
Make braces optional for single-line if, do, while, and for

gerkey:
Our style guide (https://github.com/ros2/ros2/wiki/Developer-Guide#c-1) starts with the Google style guide and then makes changes. The Google style guide says that braces are optional for [single-line if statements](https://google.github.io/styleguide/cppguide.html#Conditionals) and also for [single-line loops](https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements). Our guide doesn't modify that stance. And I like foregoing the braces when they're not needed.

This PR brings our configuration into agreement with our guide.

### DIFF
--- a/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
+++ b/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
@@ -1514,16 +1514,16 @@ nl_between_annotation                     = ignore   # ignore/add/remove/force
 #
 
 # Add or remove braces on single-line 'do' statement
-mod_full_brace_do                         = force    # ignore/add/remove/force
+mod_full_brace_do                         = ignore    # ignore/add/remove/force
 
 # Add or remove braces on single-line 'for' statement
-mod_full_brace_for                        = force    # ignore/add/remove/force
+mod_full_brace_for                        = ignore    # ignore/add/remove/force
 
 # Add or remove braces on single-line function definitions. (Pawn)
 mod_full_brace_function                   = force    # ignore/add/remove/force
 
 # Add or remove braces on single-line 'if' statement. Will not remove the braces if they contain an 'else'.
-mod_full_brace_if                         = force    # ignore/add/remove/force
+mod_full_brace_if                         = ignore    # ignore/add/remove/force
 
 # Make all if/elseif/else statements in a chain be braced or not. Overrides mod_full_brace_if.
 # If any must be braced, they are all braced.  If all can be unbraced, then the braces are removed.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this is a good idea.

In C / C++ it is easy to run into issues when not using curly braces on single line statement, e.g. when calling a macro. If the macro does some magic and optimizes itself out in release mode your condition suddenly affects the new lines. The potential for such issues is usually the reason why curly braces are enforced also for single-line statements.

Also since we don't have newlines before curly braces we don't loose much vertical space by having them.

gerkey:
I see the risk, but disagree.

wjwwood:
I agree that we shouldn't allow braces-less blocks, see: https://www.cigital.com/blog/understanding-apple-goto-fail-vulnerability-2/

dirk-thomas:
Another example against:

```
if (cond1)
   if (cond2) 
      doSomething();
else
   doSomethingElse();
```

gerkey:
Other opinions on this?

tfoote:
I'd prefer to always use the braces too. It takes a little bit more screen space but allows for quicker scanning when reading the code.

gerkey:
I still disagree, but there are more of you than there are of me.

dirk-thomas:
Since you brought this up I think it still needs to be acted upon. This is just one of the many cases where we blindly refer to the Google style guide but are actually not following it. We need to clarify this in the developer guide.

gerkey:
@dirk-thomas Good point. I [added it to the guide](https://github.com/ros2/ros2/wiki/Developer-Guide/_compare/c776fe7c61b091c02b946f8ef39deaf9521e72b7...84652f8af31db3af2e50a3ac69ee571a6a3ab679).

Let's keep adding these divergences to the guide as we discover and discuss them. I find it very unpleasant to read the guide, have all the rules in my head, do the right thing, and then still be told that I'm wrong by the unfeeling beast that is uncrustify/cpplint.

dirk-thomas:
I think adding them as we discover them is possible but also a very long process. There are numerous cases where we diverge. Several we already know of and many more we haven't even discovered / don't remember that they are actually in there.

When we talked about these in general I made the point that we should not blindly reference them because for many things we don't even know what's mentioned in the doc. We should revisit this after this cycle in a meeting.

gerkey:
@dirk-thomas Sure, let's discuss it further. I can appreciate that fully describing our style guide is difficult. But we should do it, to arrive at a narrative explanation of what we do, and why.

I'm not satisfied telling new developers: "The required style is whatever the uncrustify and cpplint configuration files contain. They are gospel and don't require any explanation. Just do what they tell you, or else you're wrong."

PR #46, opened 2016-02-19
fix algo to determine --root for cpplint

dirk-thomas:
For a path `/somewhere/src/other/test/file.hpp` it picks `/somewhere/src` as root without this patch.

With this patch it uses the correct root for `rcl/test` which lead to a new style warning (http://ci.ros2.org/job/ci_linux/1022/) which is addressed by the referenced PR (http://ci.ros2.org/job/ci_linux/1023/).


jacquelinekay:
+1

PR #47, opened 2016-03-25
ignore pep257 error D203 by default

wjwwood:
I don't know why this one is getting flagged on my machine and CI (maybe our version of pep257 is out-of-date, but I couldn't find a commit where this was changed) because their documentation explicitly says that it is not in the default list:

From http://pep257.readthedocs.org/en/latest/error_codes.html#default-checks:
> All of the above error codes are checked for by default except for D203.

Either way, I think we should explicitly ignore it by default as well.

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -38,7 +38,7 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--ignore',
         nargs='*',
-        default=['D100', 'D101', 'D102', 'D103'],
+        default=['D100', 'D101', 'D102', 'D103', 'D203'],
         help='The pep257 categories to ignore')
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


wjwwood:
Here's an example of it failing on the farm: http://ci.ros2.org/job/ci_linux/1115/testReport/rclpy/pep257/D203____rclpy_exceptions_py_16_/

wjwwood:
Also, I should mention that after perusing through https://www.python.org/dev/peps/pep-0257/ I can't find anywhere that they even suggest a new line before doc strings in a class. This is the closest thing I could find about it:

> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

dirk-thomas:
LGTM

PR #49, opened 2016-04-26
use upstream pydocstyle

dirk-thomas:
This should only be merged once we switched to Xenial.


dirk-thomas:
This still needs review.

wjwwood:
+1

ament_pep257/ament_pep257/main.py
```diff
@@ -98,21 +98,30 @@ def main(argv=sys.argv[1:]):
 
 
 def generate_pep257_report(paths, excludes, ignore):
-    def match(filename):
-        return filename.endswith('.py')
-
-    def match_dir(dirname):
-        return dirname[0] not in ['.', '_']
-
-    files = collect(paths, match=match, match_dir=match_dir)
```
wjwwood: Looks like the `paths` variable is no longer being considered.

PR #51, opened 2016-05-07
extend linter API to allow overriding the max line length

dirk-thomas:
Code generators can't always ensure that the generated code stays within a specific maximum line length. So using this that setting can be overridden.

### DIFF
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -17,9 +17,11 @@
 from __future__ import print_function
 
 import argparse
+from configparser import ConfigParser
 import difflib
 import filecmp
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -46,6 +48,9 @@ def main(argv=sys.argv[1:]):
         dest='config_file',
         help='The config file')
     parser.add_argument(
+        '--linelength', metavar='N', type=int,
+        help='The maximum line length (default: specified in the config file)')
+    parser.add_argument(
         'paths',
         nargs='*',
         default=[os.curdir],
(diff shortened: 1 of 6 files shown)


ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -72,6 +77,20 @@ def main(argv=sys.argv[1:]):
...
+        config = ConfigParser()
+        with open(args.config_file, 'r') as h:
+            config_str = h.read()
+        config.read_string('[DEFAULT]\n' + config_str)
+        code_width = config['DEFAULT']['code_width']
+        code_width = int(re.split('[ \t#]', code_width, maxsplit=1)[0])
+        if args.linelength != code_width:
+            # generate temporary config file with custom line length
+            temp_config = tempfile.NamedTemporaryFile('w')
+            temp_config.write(config_str + '\ncode_width=%d' % args.linelength)
```
tfoote: This appends the config parameter, but does not remove the original if it's already set. Will that work correctly in that case?

dirk-thomas: Yes, in an ini-like file the last value is the one being used.

tfoote:
+1

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -72,6 +77,20 @@ def main(argv=sys.argv[1:]):
...
+        # check if different from config file
+        config = ConfigParser()
+        with open(args.config_file, 'r') as h:
+            config_str = h.read()
+        config.read_string('[DEFAULT]\n' + config_str)
+        code_width = config['DEFAULT']['code_width']
+        code_width = int(re.split('[ \t#]', code_width, maxsplit=1)[0])
+        if args.linelength != code_width:
+            # generate temporary config file with custom line length
+            temp_config = tempfile.NamedTemporaryFile('w')
```
dirk-thomas: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

Fixed in a5c857c4e1ca8f89a835dc7559d3c371b14fa596.

PR #52, opened 2016-05-11
fix handling of --root

dirk-thomas:
If the `RepositoryName` does not start with the prefix also try if the full `filename` path does. This is the case when a custom absolute `--root` is being passed.

Fixes e.g. http://ci.ros2.org/job/ci_linux/1272/testReport/junit/test_communication/cpplint_rosidl_generated_cpp/build_header_guard__5____home_rosbuild_ci_scripts_ws_build_test_communication_rosidl_generator_cpp_test_communication_msg_dynamic_array_primitives__struct_hpp_303_/

After: http://ci.ros2.org/job/ci_linux/1279/

Connect to ros2/rosidl#130

### DIFF
--- a/ament_cpplint/ament_cpplint/main.py
+++ b/ament_cpplint/ament_cpplint/main.py
@@ -48,7 +48,12 @@ def CustomGetHeaderGuardCPPVariable(filename):
         # use consistent separator on Windows
         if os.sep != '/':
             prefix = prefix.replace(os.sep, '/')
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
lgtm +1

dirk-thomas:
Added 45e58126d9aaa7d6bcc91b1ae8d99000b75b9797 to fix an undefined variable in case of early exception (e.g. when passing not existing file names).

PR #54, opened 2016-06-10
update cpplint

dirk-thomas:
Since upstream has modified their repository root algorithm the second commit is necessary to find the vcs root with the longest path (see google/styleguide#145).

http://ci.ros2.org/job/ci_osx/1114/

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -28,7 +28,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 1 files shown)


wjwwood:
+1

wjwwood:
I think this cpplint failure is a regression of this:

http://ci.ros2.org/job/ci_linux/1439/testReport/rcl/cpplint/build_include__5____home_rosbuild_ci_scripts_ws_src_ros2_rcl_rcl_src_rcl_common_c_20_/

Any idea how to address that? I don't think including it from `rcl/src/rcl/common.h`, as cpplint suggests, is the right thing to do. Should I just disable linting on that line?

wjwwood:
Similar regression here: http://ci.ros2.org/job/ci_osx/1127/testReport/rosidl_generator_c/cpplint/build_include__5____Users_osrf_jenkins_workspace_ci_osx_ws_src_ros2_rosidl_rosidl_generator_c_test_separate_compilation_c_15_/

dirk-thomas:
Suppressed in ros2/rcl@10c14e02e0c3ad87f020a02fafd835ff5b92b18b and ros2/rosidl@461b4d5d6cb4caf6a21354e7276b47cd58fe231a

wjwwood:
Thanks.

PR #55, opened 2016-06-10
update clang format

dirk-thomas:
Split into Python and CMake package as well as use a configuration file to define the style.

Currently the config uses the Google style and is not matching our existing code.


ament_clang_format/ament_clang_format/main.py
```diff
@@ -44,7 +72,7 @@ def main(argv=sys.argv[1:]):
         print('No files found', file=sys.stderr)
         return 1
 
-    bin_names = ['clang-format-3.5', 'clang-format-3.4', 'clang-format-3.3']
+    bin_names = ['clang-format-3.8', 'clang-format-3.7', 'clang-format-3.6', 'clang-format-3.5', 'clang-format-3.4', 'clang-format-3.3']
```
wjwwood: On OS X the bin name is just `clang-format`.

ament_clang_format/ament_clang_format/main.py
```diff
@@ -54,7 +82,10 @@ def main(argv=sys.argv[1:]):
     report = []
 
     # invoke clang_format
-    style = '{BasedOnStyle: Google}'
+    with open(args.config_file, 'r') as h:
+        content = h.read()
+    data = yaml.load(content)
+    style = yaml.dump(data, default_flow_style=True, width=float('inf'))
```
wjwwood: Does this work? I was playing with reviving this package a couple of weeks ago, but I was unable to get into a situation where it would take my configuration file, it's really dumb that you can't just give it a file here... So you're dumping the configuration to a string and passing the whole thing as an argument?

dirk-thomas: Yes.

wjwwood:
Seems reasonable to me, though I didn't check to see if it matched our style. The other thing I noticed when trying to revive this before is that the setting may not be flexible enough to meet our current style. I don't remember off hand which thing was giving me trouble, but there were a few things you couldn't seem to change.

If the goal is just to get it working but not have it match our current style then +1.

ament_clang_format/ament_clang_format/main.py
dirk-thomas: Added in abb625f9dda28089c0af361fed41f8e179f32bb2

dirk-thomas:
This patch gets it in a working and configurable state. based on that anyone who would like to fiddle with the style options can try to update the config to match our style (or our style updated to match clang-format).

PR #58, opened 2016-09-30
(fix) correct pydocstyle import for version 1.1.0

Karsten1987:
as for version 1.1.0 pydocstyle changed their file organization. Thus, the imports have to be adjusted.

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -25,9 +25,16 @@ from xml.sax.saxutils import escape
 from xml.sax.saxutils import quoteattr
 
 from pydocstyle import check
-from pydocstyle import ConfigurationParser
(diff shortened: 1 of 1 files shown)


ament_pep257/ament_pep257/main.py
```diff
@@ -25,9 +25,22 @@
...
+try:  # as of version 1.1.0
+    from pydocstyle.violations import Error
+except ImportError:  # try version 1.0.0
+    from pydocstyle import Error
+
+try:  # as of version 1.1.0
+    from pydocstyle.utils import log
+except ImportError:  # try version 1.0.0
+    from pydocstyle import log
+
```
dirk-thomas: You can put all new imports in a single try block and all old imports in the single except block.

Karsten1987:
updated

PR #61, opened 2016-09-30
(dev) enforcing multiline docstring format

Karsten1987:
by ignoring D212 we implicitely enforce D213 as the multistring comment
format

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -45,7 +45,7 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--ignore',
         nargs='*',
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Based on http://pep257.readthedocs.io/en/latest/error_codes.html#default-checks and http://ci.ros2.org/job/ci_linux/1764/testReport/test/test_pep257/ I would say `D404` should be ignored by default, too.

dirk-thomas:
Please do not change the file `cpplint.py`. It is imported from upstream and doesn't need to comply to our rules. It should just be excluded from the linting step (which it should alread be https://github.com/ament/ament_lint/blob/845ecf69fe30c4ecff75f73497132044a58dba1b/ament_cpplint/test/test_pep257.py#L19).

Issue #53, opened 2016-05-24
Python module import order not being checked


dhood:
If we start using flake8 we can use the flake8-import-order plugin as suggested [here](https://github.com/PyCQA/pycodestyle/issues/329)

wjwwood:
+1 for `flake8`. In the past I suggested it, but I was met with resistance. If I remember correctly something about selecting `pep8` without `pyflakes`, but honestly I think it was just because we already had `pep8` and `pyflakes` separately. I think `flake8` is just superior as it has more recent development and newer options. I'd help test and review if you wanted to create a flake8 linter to replace the pep8 and pyflakes linters.

mikaelarguedas:
Note: just saw that flake8 was absent from install instructions from linux (while in the OSX and Windows instructions). Given that flake8 is installed via pip on all other platform should we add it to the packages install via pip on the linux instructions page?

codebot:
ideas to move forward:
* detect if it's flake8-v2 and disable (at very least) rather than blow up
* on farm, remove from apt line and add to pip to get flake8-v3

mikaelarguedas:
closing this now that all PRs have been merged

PR #63, opened 2016-10-18
Test python module import order using flake8

dhood:
connects to ament/ament_lint#53
requires https://github.com/ros2/ci/pull/3

This currently adds flake8 as a linter which has pep8 and pyflakes checks included by default. I have added the flake8-import-order plugin for flake8 in https://github.com/ros2/ci/pull/3.

This can replace the `ament_pep8` and `ament_pyflakes` linter packages if appropriate, and with the pep257 plugin it can replace the `ament_pep257` package as well. Should I go ahead and do that?

[Build Status](http://ci.ros2.org/job/ci_linux/1783/)
These are the files that are failing the new linter test: http://ci.ros2.org/job/ci_linux/1783/testReport/ using the [Google style guide](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_google.py) ([some](http://ci.ros2.org/job/ci_linux/1783/testReport/rosidl_parser/flake8/I100____test_test_parse_primitive_value_string_py_19_1_/) will be fixed by https://github.com/PyCQA/flake8-import-order/pull/94)


ament_cmake_flake8/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+cmake_minimum_required(VERSION 2.8.3)
```
dirk-thomas: We require version 3.5.

ament_cmake_flake8/ament_cmake_flake8-extras.cmake
```diff
@@ -0,0 +1,22 @@
+# Copyright 2014-2015 Open Source Robotics Foundation, Inc.
```
dirk-thomas: The copyright year should be 2016.

Same below

ament_cmake_flake8/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>ament_cmake_flake8</name>
+  <version>0.0.0</version>
+  <description>
+    The CMake API for ament_flake8 to check code syntax and style conventions
+    with flake8.
+  </description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
```
dirk-thomas: This should be your name.

Same below.

dhood:
comments addressed in https://github.com/ament/ament_lint/pull/63/commits/ee080bcbc80330db82ec357fe746dbcdaf7b8897. replace pep8, pyflakes, and pep257?

ament_lint_common/package.xml
```diff
@@ -16,6 +16,7 @@
   <exec_depend>ament_cmake_cppcheck</exec_depend>
   <exec_depend>ament_cmake_cpplint</exec_depend>
   <exec_depend>ament_cmake_lint_cmake</exec_depend>
+  <exec_depend>ament_cmake_flake8</exec_depend>
```
dirk-thomas: Please move this line one up to be in alphabetical order.

dirk-thomas:
Has anyone successfully setup this linter in the editor? I have `pycodestyle`, `pydocstyle`, and `pyflakes` working but not yet `flake8`.

wjwwood:
I only use flake8 in Sublime.

wjwwood:
Sorry, hit send on accident. I use this plugin:

https://github.com/SublimeLinter/SublimeLinter-flake8

I think there is an atom one too:

https://atom.io/packages/linter-flake8

I haven't tried vim or emacs.

dirk-thomas:
I had to configure the Atom package `linter-flake8` to use the `flake8` binary installed via pip (since the Ubuntu package is too old to support *on-the-fly* linting (while you type, no need to save the file, passing the content via stdin).

ament_flake8/ament_flake8/main.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class CustomStyleGuide(flake8.engine.NoQAStyleGuide):
+
+    def input_file(self, filename, **kwargs):
+        self.options.reporter.files.append(filename)
+        return super(CustomStyleGuide, self).input_file(filename, **kwargs)
+
+
+class CustomReport(pep8.StandardReport):
```
dirk-thomas: Is there a way to implement this without using `pep8` directly (which the package doesn't declare a dependency on)?

dhood:
following from https://github.com/ros2/ci/pull/3#discussion_r83973284: this uses flake8 v2.x. If not using flake8 3.x instead of 2.x is a blocker, I'll put this on the backburner since it'll likely take me a bit of effort to get it to work with 3.x (and this isn't particularly high priority)

FYI I saw your issue/PR on flake8-import-order re the import order capitalisation sensitivity of the google style guide: I had already opened https://github.com/PyCQA/flake8-import-order/pull/94 to address it (it was referenced in this PR but I could have been clearer)

ament_flake8/ament_flake8/main.py
dirk-thomas: dhood/ament_lint#1 gets rid of the `pep8` usage.

dhood:
OK, this should be good for review again please
[Build Status](http://ci.ros2.org/job/ci_linux/2181/)

dhood:
Are we interested in any of the other flake8 plugins [in this list](https://pypi.python.org/pypi?%3Aaction=search&term=flake8-&submit=search)?

dirk-thomas:
What is the fallback behavior if only version 2 of `flake8` is available since that will be the case when we create Debian packages on Xenial at some point?

dirk-thomas:
> Are we interested in any of the other flake8 plugins [in this list](https://pypi.python.org/pypi?%3Aaction=search&term=flake8-&submit=search)?

Several of these plugins look interesting. I would suggest to get the current minimum set merged before considering to add more though.

dhood:
with only flake8 v2.x the test will be added, but will fail (implemented in https://github.com/ament/ament_lint/pull/63/commits/81e6f102363c8e5de9866b10bd18bd191dfff601)

I felt it would be misleading if the test still passes.

Would it make sense to add a version check in the hook to stop it from getting added in that case? https://github.com/ament/ament_lint/pull/63/files#diff-da8d8d587f53d8cedf3b2cd20757fc40R16
I thought it might not make sense, since setups might be different when tests are added vs run.

Alternatively, I did have it working with v2.x originally, so we can choose to support both

dhood:
there's also the option of adding a `flake8>3.0` requirement to the setup.py of `ament_flake8` but since other ament linters don't do this I figured we still want to be able to install the package even without flake8 installed at all

dirk-thomas:
Once `flake8` provides `pep8` / `pep257` linting we will likely remove those from `ament_lint_common`. At that point I think `flake8` should still provide that functionality even if it is "only" run on Xenial with Debian dependencies installed (which implies `flake8` 2.x). It will only not provide the include-order checks but that is an acceptable fallback in my opinion. Therefore I think the package should gracefully fall back and support older `flake8` versions.

dhood:
ok https://github.com/ament/ament_lint/pull/63/commits/bf07f5b99ebcab3f0fa8abe7a807458ca327afb7 adds support for both.

The tests will fail with v3, however, because the `--cover-inclusive` flag that gets [set for all `ament_python` tests](https://github.com/ament/ament_tools/blob/66616f55215c4b6f5303eda772fd803d1b4e6fbc/ament_tools/build_types/ament_python.py#L116) causes the legacy file to be run, so I'm still working on a fix for that.

dhood:
tests can run if https://github.com/ament/ament_tools/pull/130 is used [Build Status](http://ci.ros2.org/job/ci_linux/2181/). That PR removes the `cover-inclusive` option but it will affect more packages than just this one, so I'm open to other (more local) suggestions for how to avoid the legacy file causing coverage tests to fail. (exceptions and sys.exit make the test fail)

Regarding building debs, here's a job with flake8 installed from apt (i.e. v2) and no import order (using https://github.com/ros2/ci/compare/flake8-apt), which I understood is what we would have for debian packages: [Build Status](http://ci.ros2.org/job/ci_linux/2193/) (notice fewer linter errors, and pip freeze lists flake8 v2.5.4):
```
00:04:17.872 ==> "/usr/bin/python3" -m pip freeze
00:04:18.243 coverage==3.7.1
00:04:18.243 empy==3.3.2
**00:04:18.243 flake8==2.5.4**
00:04:18.244 funcsigs==0.4
00:04:18.244 mccabe==0.2.1
00:04:18.244 mock==1.3.0
00:04:18.244 nose==1.3.7
00:04:18.244 pbr==1.8.0
00:04:18.244 pep8==1.7.0
00:04:18.244 pexpect==4.2.1
00:04:18.244 ptyprocess==0.5.1
00:04:18.244 pydocstyle==1.0.0
00:04:18.244 pyflakes==1.1.0
00:04:18.244 PyYAML==3.11
00:04:18.244 six==1.10.0
00:04:18.244 vcstool==0.1.21
00:04:18.244 virtualenv==15.1.0
```
so I think it should be good for debian packaging.

dhood:
I'll just merge the `legacy.py` file causing issues with the coverage job into `main.py`. The file will be a bit messier but it won't require changes to how coverage is run on ament_python packages (still open to ideas for alternatives if anyone has run into this issue before)

dirk-thomas:
Merging those two files will make it more effort to remove the legacy code in the future. Is there really no way to preserve the separation?

dhood:
OK I've fixed it. rather than trying to prevent the file from being imported, I just prefixed the whole file with `if LooseVersion(flake8.__version__) < '3.0':` (https://github.com/ament/ament_lint/pull/63/commits/036f3c2b5534f3444bf14c63238c51e0a00e34a6#diff-aa7c672db8ce0dbbacbace23d3cbd10eR23). So this PR no longer requires https://github.com/ament/ament_tools/pull/130

these are the normal jobs (using https://github.com/ros2/ci/pull/3, which gets flake8 v3, from pip):
Linux: http://ci.ros2.org/job/ci_linux/2223/#showFailuresLink (25 failed tests)
OS X: http://ci.ros2.org/job/ci_osx/1685/testReport/ (`rttest` and `pendulum_control` skipped (-4) + 2 non-lint test failures = 23)
Windows: http://ci.ros2.org/job/ci_windows/2212/testReport/ (`rttest` and `pendulum_control` skipped (-4) + 1 non-lint test failures = 22)

this is linux using flake8 from apt (v2) and import-order from pip (same result as above):
http://ci.ros2.org/job/ci_linux/2227

and this is what it would be like if we are using flake8 v2 and no import-order plugin (as would be the case for debians) - no issues if the plugin is not found:
http://ci.ros2.org/job/ci_linux/2224/testReport/

I have matching PRs for the linter fixes that I'll open once this gets approved.

mikaelarguedas:
Any reason to delay the PRs with the linter fixes until this one gets approved?

dhood:
I figured that depending on the outcome of this PR, the changes required to make linters pass for other packages might change

dirk-thomas:
This looks goo to me. With another round of CI which uses your linter fixes this should be good to go.

dhood:
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1702/)
Linux (no linter failures): [Build Status](http://ci.ros2.org/job/ci_linux/2249/)
Windows (no linter failures): [Build Status](http://ci.ros2.org/job/ci_windows/2228/)

mikaelarguedas:
Can be done in a follow up but it would be great to run flake8 on `ament_python` packages as well

ament_lint_cmake/package.xml
```diff
@@ -11,6 +11,7 @@
   <license>Apache License 2.0</license>
 
   <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
```
dirk-thomas: Are there any cases which are covered by pep8 / pep257 / pyflakes which are not covered by flake8? If not I would suggest to remove the redundant tests (ament_pep8, ament_pep257, ament_pyflakes).

ament_lint_common/package.xml
```diff
@@ -16,6 +16,7 @@
   <exec_depend>ament_cmake_copyright</exec_depend>
   <exec_depend>ament_cmake_cppcheck</exec_depend>
   <exec_depend>ament_cmake_cpplint</exec_depend>
+  <exec_depend>ament_cmake_flake8</exec_depend>
```
dirk-thomas: Are there any cases which are covered by pep8 / pep257 / pyflakes which are not covered by flake8? If not I would suggest to remove the redundant tests (ament_pep8, ament_pep257, ament_pyflakes).

dhood: It should cover `pep8` and `pyflakes`, yes. I'll remove `ament_pep8` and `ament_pyflakes` from `ament_lint_common` so that they are not added when `ament_lint_auto` is used.

I haven't added the docstrings flake8 plugin. Since you mentioned the plugins won't be installable when we build debians, perhaps that's a reason to leave `ament_pep257` as is - what do you think?

dirk-thomas: Sounds good to me.

ament_flake8/ament_flake8/main.py
```diff
@@ -0,0 +1,264 @@
...
+    print('')
+    print('%d files checked' % len(report.files))
+    if not report.total_errors:
+        print('No errors or warnings')
+        rc = 0
+    else:
+        print('%d errors' % (report.total_errors))
+
+        print('')
+        error_type_counts = report.get_error_type_counts()
```
dirk-thomas: This line fails when the legacy API is being used since [this class](https://github.com/ament/ament_lint/blob/16cb042e5d04e6d50b60f4c5d373d356052370d9/ament_flake8/ament_flake8/legacy.py#L27) doesn't have that method.

dhood: thanks for the heads up. I'll fix it

ament_flake8/ament_flake8/main.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def generate_flake8_report(config_file, paths, excludes, max_line_length=None):
+    if LooseVersion(flake8.__version__) < '3.0':
+        from ament_flake8.legacy import generate_flake8_report
+        return generate_flake8_report(config_file, paths, excludes, max_line_length)
+
+    flake8_argv = []
+    flake8_argv.append('--config={0}'.format(config_file))
+    flake8_argv.append('--exclude={0}'.format(excludes))
```
dirk-thomas: This line doesn't work. See #77

PR #67, opened 2016-12-01
cpplint: allow using-directive for user-defined literals namespaces

Sarcasm:
As proposed here:

- https://github.com/ros2/rclcpp/pull/284#issuecomment-264165706

Related to ros2/rclcpp#284.


wjwwood:
Thanks, we'll see if we end up needing this or not.

I removed the "Connects to" because this will not be needed for the other pr's since you added the nolint lines.

dirk-thomas:
For completeness I will repeat the comment we received from upstream about considering to propose this change to them:

> We (internally) had a long discussion recently and came to the conclusion that using directives are never ever ever ever safe for use in a codebase of any significant size - we cannot endorse this, even just once. > Sorry.
>
> (I'm working on finding a proper forum to share some of the Google-internal guidance, hopefully early next year.)
>
> -- https://github.com/google/styleguide/pull/204#issuecomment-264502185

wjwwood:
Anyone following this should read also the follow up comment specifically about `using namespace std::*literals`:

https://github.com/google/styleguide/pull/204#issuecomment-265744834

ament_cpplint/ament_cpplint/cpplint.py
```diff
@@ -4736,11 +4736,14 @@ def CheckLanguage(filename, clean_lines, linenum, file_extension,
...
   match = Search(r'\busing namespace\s+((\w|::)+)', line)
-  if (match and
-      (IsHeaderExtension(file_extension) or 'literals' not in match.group(1))):
-    error(filename, linenum, 'build/namespaces', 5,
-          'Do not use namespace using-directives.  '
-          'Use using-declarations instead.')
+  if match:
+    matched = match.group(1)
+    match_contains_std_literals = \
+      matched.startswith('std::') and matched.endswith('literals')
```
Sarcasm: If restricting to the STL is decided, then the comments should also be updated.

Sarcasm:
Hello @codebot,

TBH, I'm glad GCC has this warning bug because I don't consider the use of the using declaration elegant (personal taste).

Regarding restricting the match to `std::` members, I'm not a big fan either but you are the judges.
In the following PR https://github.com/ros2/rclcpp/pull/284,
I'm adding a user defined literal namespace to rclcpp.
I'm trying to follow the STL convention of a namespace named with `literals`, which I think is a convention quite a few people would follow naturally. And estimate the number of `literals` namespace that will be abused by people in the ros2 codebase without people noticing in the review to be very close to 0 (risky estimate). Personally I think UDL are nice, for test code, examples, a bit less on real-world code, and I don't think the use should be limited to the STL one, if Boost, or any other library defines some, it makes sense to allow them when then follow the STL pattern of naming the namespace `*literals`.

I haven't given extensive thoughts on the subject but I'm wondering if the `std::placeholders` namespace could not be allowed as well.

Regarding the flexibility of the linting tool.
I'm wondering if `clang-tidy` (http://clang.llvm.org/extra/clang-tidy/checks/google-build-using-namespace.html) would not be more welcoming to configurable rules that diverges from the Google Style.
But I digress, clang-tidy integration would be significantly more difficult to integrate than cpplint.

Sarcasm:
:+1:

ament_cpplint/ament_cpplint/cpplint.py
```diff
@@ -4729,10 +4729,31 @@ def CheckLanguage(filename, clean_lines, linenum, file_extension,
...
+  if match:
+    whitelist = [
+      'std::chrono_literals',
+      'std::complex_literals',
+      'std::literals',
+      'std::string_literals',
+      'std::literals::chrono_literals',
+      'std::literals::complex_literals',
+      'std::literals::string_literals',
+      'std::placeholders',
```
dirk-thomas: Maybe order the list?

wjwwood:
@codebot I added a few literals in https://github.com/ament/ament_lint/pull/67/commits/ea5846f310f4116ea8035280a5444719b052cfe5 and fixed the alphabetical ordering in https://github.com/ament/ament_lint/pull/67/commits/787ae55d0fea4b2d19dfcd533e50326926685e89 (doh)

wjwwood:
As much as I hate to have patches to cpplint in our project, I think this pr represents the most pragmatic compromise between using new features in the standard as they were intended versus avoiding the abuse of the `using` directive. So, +1 from me.

@dirk-thomas didn't we have a place to keep track of the changes we have on top of cpplint? I saw the url at the top of the file to the originating commit on the cpplint repository, but not an abbreviated list of changes or anything.

dirk-thomas:
I don't think such a list exists anywhere beside the commit log (https://github.com/ament/ament_lint/commits/master/ament_cpplint/ament_cpplint/cpplint.py). All customizations are can be found in the commit after the last import from upstream (https://github.com/ament/ament_lint/commit/232428752251de61e84ef013bcd643e35eb9038d#diff-b5c57d8cc3ce83e990089f94f5ab6e15).

PR #69, opened 2017-02-28
Remove xml warning

mikaelarguedas:
This PR uses xml version 2 for cppcheck output.

Rationale:
cppcheck deprecated usage of version 1 in the last releases, and will completely drop support in v1.81.
We started to see deprecation warnings show on Mac and Windows CIs (`cppcheck: XML format version 1 is deprecated and will be removed in cppcheck 1.81. Use '--xml-version=2'`) e.g. http://ci.ros2.org/view/nightly/job/nightly_osx_debug/373

Tested with a few errors for either same or different kind and the generated xunit file is identical as with version 1.
Errors tested:
```cpp
void f(char *p) {
    if (p+1){}
}

bool using_introspection_c_typesupport(const char * typesupport_identifier)
{
  double * a = reinterpret_cast<double *>(malloc(10 * sizeof(double)));
  (void)a;
  return true;
}

```

### DIFF
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -79,7 +79,7 @@ def main(argv=sys.argv[1:]):
            '-q',
            '-rp',
            '--xml',
-           '--xml-version=1']
+           '--xml-version=2']
     if jobs:
         cmd.extend(['-j', '%d' % jobs])
     cmd.extend(files)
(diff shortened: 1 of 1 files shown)


ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -102,13 +102,15 @@ def main(argv=sys.argv[1:]):
     report = {}
     for filename in files:
         report[filename] = []
-    for error in root.findall('error'):
-        filename = error.get('file')
+    errors = root.find('errors')
+    for error in errors:
```
dirk-thomas: Since `errors` is only used once this could stay in one line as before: `for error in root.find('errors'):`

mikaelarguedas: fixed in https://github.com/ament/ament_lint/pull/69/commits/c279be12d5366ee1045e2276a2c5e6900a71d677

Issue #72, opened 2017-04-25
Add in support for the BSD license in the copyright linting


clalancette:
Some of the code that is coming into the codebase originally came from ROS1 with a BSD license.  Since we are not making substantial changes to it, we can't really change the license (see https://github.com/ros2/common_interfaces/commit/025b1911ab2b135ea009017b06069a88e7e05817#commitcomment-21888534 for a discussion on this).

However, we would like to lint these files.  To do this, I think we'll have to add support for the BSD copyright into ament_lint.  For an example of it failing, see: http://ci.ros2.org/job/ci_linux/2531/ , where the failing file is https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/include/sensor_msgs/distortion_models.hpp

mikaelarguedas:
For reference this related to https://github.com/ament/ament_lint/pull/66 and https://github.com/ament/ament_lint/issues/65

dirk-thomas:
You can add new licenses by registering a new license as an entry point: https://github.com/ament/ament_lint/blob/6c7a0ab90b8c620d4e6f28cb5a56254f065f48a1/ament_copyright/setup.py#L34

I don't think it is related to the referenced tickets since you want it to support BSD - not packages with multiple licenses.

clalancette:
Closing since #73 was closed.

PR #73, opened 2017-04-26
Bsd2 license

clalancette:
This PR adds support for the BSD license into ament_lint.  This is necessary for turning on all linting in sensor_msgs, since some of that code is under the BSD 2 license.

fixes ament/ament_lint#72

CI:
windows: [Build Status](http://ci.ros2.org/job/ci_windows/2645/)
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2542/)
linux aarch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/141/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2002/)


clalancette:
In terms of review, this is probably easiest to review patch-by-patch, rather than the combined diff (particularly for the regex changes).

ament_copyright/ament_copyright/parser.py
```diff
@@ -166,7 +166,7 @@ def search_copyright_information(content):
     year = '\d{4}'
     year_range = '%s-%s' % (year, year)
     year_or_year_range = '(?:%s|%s)' % (year, year_range)
-    pattern = '^[^\n\r]?\s*Copyright\s+(%s(?:,\s*%s)*)\s+([^\n\r]+)$' % \
+    pattern = '^[^\n\r]?\s*Copyright\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
```
dirk-thomas: Can you give an example what wasn't accepted before and is accepted with this change.

mikaelarguedas:
Do these CI jobs test any code that has a BSD licence ? e.g https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/include/sensor_msgs/distortion_models.hpp

ament_copyright/ament_copyright/parser.py
```diff
@@ -166,7 +166,7 @@ def search_copyright_information(content):
     year = '\d{4}'
     year_range = '%s-%s' % (year, year)
     year_or_year_range = '(?:%s|%s)' % (year, year_range)
-    pattern = '^[^\n\r]?\s*Copyright\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
+    pattern = '^[^\n\r]?\s*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
```
dirk-thomas: Do we have to support this case or could we update the code using that.

ament_copyright/ament_copyright/parser.py
```diff
@@ -54,14 +55,6 @@ def read(self):
...
         raise NotImplemented()
 
-    def identify_license(self, content, license_part):
-        for name, license in get_licenses().items():
-            if content is not None and getattr(license, license_part) == content:
-                self.license_identifier = name
-                break
-        else:
-            self.license_identifier = UNKNOWN_IDENTIFIER
-
```
dirk-thomas: Why is this function being removed and the logic being duplicated into multiple locations?

ament_copyright/ament_copyright/parser.py
clalancette: Sure.  Previous to this regex change, this was accepted:

Copyright 2017 Open Source Robotics Foundation, Inc.

But this wasn't:

Copyright 2017, Open Source Robotics Foundation, Inc.

(note the comma after the year).  It is fairly minor, but I think it makes this regex a little more forgiving of the format.

ament_copyright/ament_copyright/parser.py
clalancette: So, this regex change is to allow a (c) after the word "Copyright".  While we could change the code that has this, the BSD license recommends this format:

Copyright (c) <year>, <rights-holder>

(https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_.28.22BSD_License_2.0.22.2C_.22Revised_BSD_License.22.2C_.22New_BSD_License.22.2C_or_.22Modified_BSD_License.22.29).
Thus, any BSD code we come across is likely to be in that format, so I thought it would be better to support that.

ament_copyright/ament_copyright/parser.py
clalancette: So, I wasn't entirely sure what to do about this.  We need to do something slightly different in this method in the subclass SourceDescriptor than we do in ContributingDescriptor and LicenseDescriptor.  Thus, my options seemed to be:

* Override it in SourceDescriptor
* Add in some special-case flag for when we wanted to do additional substitutions
* Copy the code

Given that it's really only 4 lines of code, I opted for the last one.  However, I'm open to other suggestions.

clalancette:
@mikaelarguedas Good point, no, that CI doesn't have any BSD code.  I've been testing locally by adding the ament_lint_auto back into sensor_msgs; I can do a CI with that instead to show this actually working.

ament_copyright/ament_copyright/parser.py
dirk-thomas: I am not sure what the override is needed for (since the Apache license has the same structure already) but I would certainly prefer the first or second option.

clalancette: I need to override identify_license() because the SourceDescriptor version of it wants to iterate over the self.copyrights list, which only exists in the SourceDescriptor subclass.  I can restore the generic version to FileDescriptor and override it in SourceDescriptor, but it seems like overkill since it is such a simple function.

clalancette:
New CI, which should test the BSD license in common_interfaces/sensor_msgs now:

windows: [Build Status](http://ci.ros2.org/job/ci_windows/2649/)
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2544/)
linux aarch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/142/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2005/)

dirk-thomas:
The replacement of `{copyright}` in the source file template is not necessary. The Apache header also contains `{copyright}` and doesn't require the expansion.

Regarding the expansion of `{company}` I would rather suggest to use a "generic" snippet (see https://opensource.org/licenses/BSD-3-Clause) which refers to "Neither the name of the copyright holder nor the names of its contributors" rather than filling in the company name.

The changes to accept `(c)` and the extra comma in the copyright line are fine.

All other changes are not necessary and I would rather avoid them if there is no good reason for them. E.g. the `FileDescriptor` doesn't have the member `license_identifier` because it never has a license. Only some of the subclasses have that attribute.

clalancette:
> Regarding the expansion of {company} I would rather suggest to use a "generic" snippet (see https://opensource.org/licenses/BSD-3-Clause) which refers to "Neither the name of the copyright holder nor the names of its contributors" rather than filling in the company name. (The template also lacks a trailing newline.)

This is the key point, I think.  If we are OK with converting all incoming BSD licenses to the generic version, then we can do away with most of the code changes here.  Are we OK with doing that?

tfoote:
We shouldn't require changing license text contents to pass the linter beyond possibly stripping trailing whitespace. We can do that for our own copyright, but should not apply changes to anything for which we don't hold the copyright.

I would recommend that we allow basically arbitrary strings in the {company} section.

dirk-thomas:
> I would recommend that we allow basically arbitrary strings in the {company} section.

Sounds good to me.

clalancette:
I just pushed a different approach to this.  I changed the BSD2 template to be the generic version, and then I used difflib to compare a ratio between the incoming text and the template.  If that is > 95%, we accept that as the license (I'm not wedded to that ratio, it just seemed somewhat reasonable).

I left the move of the identify_copyright() function, and the consolidation of self.license_identifier into FileDescriptor().  Here are my reasons:

1.  identify_copyright() operates on data only present in the SourceDescriptor subclass.  Thus, it doesn't belong in a generic super class.
2.  self.license_identifier is set by all of its subclasses.  This suggests that it is a property that they all share, and thus it makes sense to me to hoist it up to the super class.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2564/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/161/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2033/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2671/)

ament_copyright/ament_copyright/parser.py
```diff
@@ -54,24 +56,17 @@ def read(self):
...
-                self.copyright_identifiers.append(UNKNOWN_IDENTIFIER)
-
     def identify_license(self, content, license_part):
+        if content is None:
+            return
+
         for name, license in get_licenses().items():
-            if content is not None and getattr(license, license_part) == content:
+            template = getattr(license, license_part).replace('\n', ' ')
+            tomatch = content.replace('\n', ' ')
```
dirk-thomas: Unused variable?

ament_copyright/ament_copyright/parser.py
```diff
@@ -44,6 +45,7 @@ def __init__(self, filetype, path):
         self.path = path
         self.exists = os.path.exists(path)
         self.content = None
+        self.license_identifier = UNKNOWN_IDENTIFIER
```
dirk-thomas: See previous comment why I don't think the base class should have that attribute.

ament_copyright/ament_copyright/parser.py
```diff
@@ -54,24 +56,17 @@ def read(self):
...
     def identify_license(self, content, license_part):
+        if content is None:
+            return
+
         for name, license in get_licenses().items():
-            if content is not None and getattr(license, license_part) == content:
+            template = getattr(license, license_part).replace('\n', ' ')
+            tomatch = content.replace('\n', ' ')
+            x = difflib.SequenceMatcher(a=template, b=content)
+            if x.ratio() > 0.95:
```
dirk-thomas: I am not sure this is "good enough". This could result in a wrong detection if he license is just similar enough.

ament_copyright/ament_copyright/template/bsd2_header.txt
```diff
@@ -0,0 +1,31 @@
...
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
```
dirk-thomas: Please see previous comment. Still lacking a newline at the end.

ament_copyright/ament_copyright/parser.py
clalancette: Oh yeah, that's a bug.  I should be using that in the difflib call.

clalancette: (it's also the cause of the test failure in the CI)

ament_copyright/ament_copyright/parser.py
clalancette: I guess, though a license that is 95% the same as this would probably be considered the same.

So the real question is whether we should just up this value (98%, say), or do you disagree with the approach?

ament_copyright/ament_copyright/template/bsd2_header.txt
clalancette: Will fix that.

dirk-thomas: "Cutting out" the company name and checking the string before and after should be straight forward. So I am not sure what the diff approach is gaining us.

clalancette: Basically, what this buys us is the ability to be a lot more forgiving with the license.

In the case of the BSD license, there are at least 3 different variations I've seen, just between looking at the common_interfaces code, the license you linked to at opensource.org, and the license on wikipedia.  In the one in the code, it has the name of the company in it, and uses * for each point made.  In the one at opensource.org, it says "copyright holder" instead of the company name, and uses numerals instead of stars for each point.  In the one on wikipedia, it has the name of the company in it, and uses * for each point.

While I could write some custom parsing code to deal with all of those differences, it would necessarily be BSD-specific, which doesn't seem like an ideal path to go down.  The difflib approach is more flexible, and should work for other licenses as well.

ament_copyright/ament_copyright/parser.py
clalancette: Right, I guess that is the point I disagree with :).  While that base class itself doesn't have a license, all of its sub-classes do (they all call identify_license()).  Thus it seems to be something common to all of them, and seems to make sense in the base class.  It's similar in many respects to the identify_license() method, in that it is a convenient place to not duplicate the code, rather than it being somehow intrinsic to the base class.

tfoote:
I don't think that this new approach is appropriate for this use case. Comparing the fraction of match is a great way to get yourself into a lot of trouble. If I add a "not" in any clause, it completely changes the meaning and compliance but does not significantly change the diff. If there's any change outside the standard parameters it's no longer the named license but a variant which needs to be interpreted as it's own entity.

clalancette:
@tfoote I thought it might be interesting to try this approach, but I agree with your assessment.  So my plan will be to split the templated license on the "{company}" string, then look for those substrings (in that order) against the incoming text to test against.  Does that sound like a better approach?

tfoote:
That sounds like a reasonable approach.

clalancette:
CI for the new approach:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2573/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/170/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2044/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2686/)

mikaelarguedas:
once this is merged https://github.com/ros2/common_interfaces/commit/a0234f2254221b9927ede3fe06dd6abc2d95559a and https://github.com/ros2/common_interfaces/pull/33 need to be reverted

clalancette:
Thanks for the review.  I know we usually squash changes together, but in this one, I'd actually prefer for it to go in as 3 different commits: one to change the regex to allow the (c), one to change the regex to allow the comma, and one to add the BSD license.  I think this will make following the regex a little easier later on.  Does that seem OK?

dirk-thomas:
Sounds good to me.

clalancette:
I've now merged this (by hand, to split the commits up).  Thanks!

dirk-thomas:
It would be better to rebase the commits on the branch of the PR. That way the actually merged changes are still visible from the PR. Currently the PR doesn't reference the actual changes done on master.

clalancette:
Oh, right, that would be better.  Should I go back and do that (force pushing master to remove the commits), or is that advice for the future?

dirk-thomas:
In this case it I would say it is too late. But you could reference the actual commit hashes which have been applied to master so that future readers find them easily.

clalancette:
This PR ended up applying these commits:

https://github.com/ament/ament_lint/commit/3f65895ec2123f7e47b98d7316866b5819984495
https://github.com/ament/ament_lint/commit/77061fd352728c3d80bb00a273b38b1788bcf1b8
https://github.com/ament/ament_lint/commit/767257ff141d7957f3c8c7b92106aced95794537

PR #76, opened 2017-06-06
Add python3 and setuptools as build dependencies.

nuclearsandwich:
In order to build python packages with pybuild python3 and setuptools must be present.

### DIFF
--- a/ament_clang_format/package.xml
+++ b/ament_clang_format/package.xml
@@ -11,6 +11,10 @@
 
   <build_depend>python3-setuptools</build_depend>
 
+  <buildtool_depend>python3</buildtool_depend>
+
+  <buildtool_export_depend>python3</buildtool_export_depend>
+
   <exec_depend>clang-format</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
--- a/ament_clang_format/package.xml
+++ b/ament_clang_format/package.xml
@@ -13,8 +13,7 @@
 
   <buildtool_depend>python3</buildtool_depend>
 
-  <buildtool_export_depend>python3</buildtool_export_depend>
-
+  <exec_depend>python3</exec_depend>
   <exec_depend>clang-format</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
--- a/ament_clang_format/package.xml
+++ b/ament_clang_format/package.xml
@@ -9,6 +9,8 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
   <exec_depend>clang-format</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
--- a/ament_clang_format/package.xml
+++ b/ament_clang_format/package.xml
@@ -13,8 +13,8 @@
 
   <buildtool_depend>python3</buildtool_depend>
 
-  <exec_depend>python3</exec_depend>
   <exec_depend>clang-format</exec_depend>
+  <exec_depend>python3</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_pep8</test_depend>
--- a/ament_cmake_cpplint/package.xml
+++ b/ament_cmake_cpplint/package.xml
@@ -9,8 +9,6 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
-  <build_depend>python3-setuptools</build_depend>
-
   <buildtool_depend>ament_cmake_core</buildtool_depend>
   <buildtool_depend>ament_cmake_test</buildtool_depend>
 
--- a/ament_cmake_cpplint/package.xml
+++ b/ament_cmake_cpplint/package.xml
@@ -9,6 +9,8 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
   <buildtool_depend>ament_cmake_core</buildtool_depend>
   <buildtool_depend>ament_cmake_test</buildtool_depend>
 
--- a/ament_copyright/package.xml
+++ b/ament_copyright/package.xml
@@ -14,7 +14,7 @@
 
   <buildtool_depend>python3</buildtool_depend>
 
-  <buildtool_export_depend>python3</buildtool_export_depend>
+  <exec_depend>python3</exec_depend>
 
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_pep8</test_depend>
--- a/ament_copyright/package.xml
+++ b/ament_copyright/package.xml
@@ -10,6 +10,8 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
   <test_depend>ament_pep257</test_depend>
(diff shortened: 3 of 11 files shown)


ament_clang_format/package.xml
```diff
@@ -9,6 +9,12 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
+  <buildtool_depend>python3</buildtool_depend>
+
+  <buildtool_export_depend>python3</buildtool_export_depend>
```
dirk-thomas: Shouldn't this be an `exec_depend` instead?

Same below.

nuclearsandwich:
If it's possible to get this in before the end of the day I won't have to rebuild its large downstream again after it lands.

With regard to the outstanding issue of whether python packages should have an explicit dependency on python: For the purposes of this release the dependency either needs to be here or needs to be patched into the package.xml at release time.

ament_clang_format/package.xml
```diff
@@ -9,6 +9,11 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
+  <buildtool_depend>python3</buildtool_depend>
+
+  <exec_depend>python3</exec_depend>
```
dirk-thomas: Should be one line below for alpha order.

ament_clang_format/package.xml
```diff
@@ -9,6 +9,11 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python3-setuptools</build_depend>
+
+  <buildtool_depend>python3</buildtool_depend>
```
dirk-thomas: Do we really need to tell a Python package that it needs `python3` to invoke the `setup.py`? It looks superfluous to me (I made the same comments on similar PRs).

nuclearsandwich: I cargo-culted this from [ament_package's package.xml](https://github.com/ament/ament_package/blame/master/package.xml) when it built and others failed.

What's really needed here in most cases is python3-dev (which is resolved by the rosdep key python3). setuptools will pull in the `python3` debian package but not python3-dev.

dirk-thomas: I don't see why `python3-dev` would be needed "in most cases".

But I don't think we should require every Python package to declare a dependency on `python3`. As mentioned before we also don't declare a dependency on `build-essential` when  building a package which needs a compiler.

nuclearsandwich: > Imo this should be handled somewhere else in the release pipeline.

Do you have a suggestion as to where? Should bloom magically inject python build deps when ament_package is the build type.
Should I just patch package.xmls at release time.

dirk-thomas: I don't have a perfect solution for it. Maybe others have ideads. But some users might use Python 3.6 and our packages would then force them to install a different Python version.

dirk-thomas: The Debian Python Policy requires a package to depend on `python3 (>= 3.Y)` (with `Y=5` in our case) (see https://www.debian.org/doc/packaging-manuals/python-policy/ch-module_packages.html#s-dependencies). So if we *would* require every Python package to declare that dependency it would for sure need to specify the correct version range.

dirk-thomas: E.g. every non-Python packages would need an explicit `cmake` dependency if we apply the same "rules". Imo both would make it much more effort for users of ROS since it is boilerplate and redundant information. It would be possible to inject this kind of dependency based on the build type. Not sure where in the process this could live best though.

nuclearsandwich:
> every non-Python packages would need an explicit cmake dependency if we apply the same "rules". > [which] would make it much more effort for users of ROS since it is boilerplate and redundant information.

Packages that require [catkin](https://github.com/ros/ros_comm/blob/ac9f45f3caa5c551f4dfeb12dee68d34cd657471/ros_comm/package.xml#L24), [ament_cmake](https://github.com/ros2/joystick_drivers/blob/22d413292862b518c1b613192213d9de27dba036/joy/package.xml#L27), or [plain cmake](https://github.com/ament/ament_cmake/blob/dc243d84429ccefbe3532c59a242a114eb20bad6/ament_cmake_core/package.xml#L22) all _do_ have explicit dependencies on those tools. The reason we need explicit dependencies on python build utilities is because there is no such package as `ament_python` for packages of that build type to depend on.

> It would be possible to inject this kind of dependency based on the build type. Not sure where in the process this could live best though.

This can be done with bloom, but not in its current state. I've been evaluating ways to improve bloom's support for alternative build types but it wasn't designed for it and hacking it even further during this release window would either completely compromise the code quality or further fragment bloom's generators meaning through copy-paste duplication.

dirk-thomas:
> This can be done with bloom, but not in its current state. I've been evaluating ways to improve bloom's support for alternative build types but it wasn't designed for it and hacking it even further during this release window would either completely compromise the code quality or further fragment bloom's generators meaning through copy-paste duplication.

As far as I can see [these two lines](https://github.com/ros-infrastructure/bloom/blob/9ea33a79fde666e9497857c0b764ffd2ff5aef76/bloom/generators/debian/generator.py#L305-L306) read the dependencies from the package manifest. Later these dependencies are being used to expand the Debian control file template (https://github.com/ros-infrastructure/bloom/blob/9ea33a79fde666e9497857c0b764ffd2ff5aef76/bloom/generators/debian/templates/control.em#L5 and https://github.com/ros-infrastructure/bloom/blob/9ea33a79fde666e9497857c0b764ffd2ff5aef76/bloom/generators/debian/templates/control.em#L11).

Since the parsed package manifest is available where the package dependencies are extracted I would expect that it should be straight forward to inject additional dependencies there based on the build type. For example with a code snippet like this:

```
build_type = package.get_build_type()
if build_type == 'ament_cmake':
    build_depends.append(Dependency('python3', version_gte='3.5'))
    depends.append(Dependency('python3', version_gte='3.5'))
```

And since the build types of ROS 1 and ROS 2 are intentionally not overlapping this could even be integrated into the current bloom code without side effects on ROS 1. Each build type could add specific information.

I certainly don't know much about the internals of `bloom` and might be missing something. Maybe you can describe why you think this can't be done "in its current state" or why it would "completely compromise the code quality". Maybe @wjwwood can clarify if the approach I described is feasible and what the downsides are.

nuclearsandwich:
> `if build_type == 'ament_cmake':`

I assume here you mean `ament_python`.


> Maybe you can describe why you think this can't be done "in its current state" or why it would "completely compromise the code quality".

We could add a condition like this, I think, the get_build_type method wasn't part of catkin_pkg when I started working on this and I vendored the code to fetch it within the method I needed it.

Bumping the minimum required version of catkin_pkg will get the method and allow adding of a conditional snippet like that. My primary hesitation is in doing so is that we're already stretching the template logic quite hard getting it to support two different build systems in the same rules file. I'll spike this out and see if there's any hangups I'm not seeing right away on a Sunday evening.

wjwwood:
> Maybe @wjwwood can clarify if the approach I described is feasible and what the downsides are.

The approach you described sounds like it will work.

I honestly don't mind which strategy you end up taking, but I don't think it would be a bad thing necessarily if all pure Python packages needed a dependency on either python2 or python3 (or maybe setuptools from one or the other). It might be tedious, but it would also be correct I think. The fact that we don't depend on build essentials or cmake in other build types has more to do with legacy and convenience than with correctness, imo.

nuclearsandwich:
Closing for now in favor of shoehorning dependencies via bloom based on the ament_python build type but I've added revisiting implicit dependencies to my list of things to discuss after the release.

PR #77, opened 2017-06-21
fix --exclude with ament_flake8

dirk-thomas:


ament_flake8/ament_flake8/main.py
```diff
@@ -139,8 +139,10 @@ def generate_flake8_report(config_file, paths, excludes, max_line_length=None):
         return generate_flake8_report(config_file, paths, excludes, max_line_length)
 
     flake8_argv = []
-    flake8_argv.append('--config={0}'.format(config_file))
-    flake8_argv.append('--exclude={0}'.format(excludes))
+    if config_file is not None:
+        flake8_argv.append('--config={0}'.format(config_file))
+    if excludes is not None:
```
dhood: Changing the check to `if not excludes:` will give consistent behaviour with the other linters in the event the list is empty.  Unless there's a reason this check is more appropriate, could you update it?

dirk-thomas: I am not sure I understand. The variable `excludes` is either `None` or a not empty list. How would it be an empty list?

dhood: From my understanding it could be an empty list if a user specifies the `--exclude` option, but with no filenames

dirk-thomas: That is not possible. `argparse` won't allow you to do that with the current settings.

Also in that case I think the code you commented on is doing exactly the right think: `excludes is not None` and therefore calls the script with `--excluded=`.

dhood: Which setting makes it not possible? From what I've seen `'--foo --bar 1 2` will make `foo` an empty list if it's an `nargs='*'` option.

My concern was consistency with the v2.x implementation, since it [doesn't use the argument if it's `[]`](https://github.com/ament/ament_lint/blob/5c9c8acaa5f0137f4544857038652c589e506a32/ament_flake8/ament_flake8/legacy.py#L76). but after re-looking at it, the end result will still turn out the same as 2.x with this implementation, so I'm fine to leave it as is

dirk-thomas: If you try to run `ament_flake8 --exclude` the result will be:

```
Usage: flake8 [options] file file ...

flake8: error: --exclude option requires 1 argument
```

dhood: oh, here's an example of how it can be an empty list:

```
In [1]: import argparse

In [2]: parser = argparse.ArgumentParser()

In [3]: parser.add_argument('--foo', nargs='*')
Out[3]: _StoreAction(option_strings=['--foo'], dest='foo', nargs='*', const=None, default=None, type=None, choices=None, help=None, metavar=None)

In [4]: parser.add_argument('--bar', nargs='*')
Out[4]: _StoreAction(option_strings=['--bar'], dest='bar', nargs='*', const=None, default=None, type=None, choices=None, help=None, metavar=None)

In [5]: parser.parse_args('--foo --bar 1 2'.split())
Out[5]: Namespace(bar=['1', '2'], foo=[])
```

it's strange that argparse has different behaviour base on its position in the list (strange to me, at least!)

PR #78, opened 2017-06-21
use flake8 instead of pep8 and pyflakes

dirk-thomas:
Connect to ament/ament_index#25.


dirk-thomas:
> I'm not sure why the copyright years were changing (maybe copy-paste?), but otherwise lgtm.

`git` thinks these files have been renamed. Instead the old files have been deleted and new files for flake8 created. The copyright year in the new files is the same as from the file they have been copied from (from other packages).

PR #80, opened 2017-06-22
use OSI website as reference for license

mikaelarguedas:
Follow up of https://github.com/ament/ament_lint/commit/767257ff141d7957f3c8c7b92106aced95794537#commitcomment-22668604

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2794/)


mikaelarguedas:
ready for review, we can hold this until after the beta if we don't have a full rebuild before then

Issue #75, opened 2017-05-16
Make ament_copyright robust to newlines


clalancette:
It should be fairly straightforward to make ament_copyright deal with newlines in a more robust way, so that lines can be broken up in different ways in the copyright header in files.

mikaelarguedas:
to be considered when implementing this: supporting block comments and thus lines starting with `*` would be valuable.
e.g.:
```c
/* Copyright notice
 * line1
 * line2
 */
```

allenh1:
@mikaelarguedas this is one I could work on, right?

mikaelarguedas:
yup! :+1:

mikaelarguedas:
@allenh1 are you still planning on working on this ?

allenh1:
Thanks for the reminder! I definitely am planning on it, and you can assign me if you want (that way I won't forget again).

allenh1:
See comments on #82.

PR #82, opened 2017-08-04
Added more robust comment format support.

allenh1:
I've tested this as-is with the following:

```c
/*
 * slam_gmapping
 * Copyright (c) 2008, Willow Garage, Inc.
 * Copyright (c) 2017, Open Source Robotics Foundation, Inc.
 *
 * THE WORK (AS DEFINED BELOW) IS PROVIDED UNDER THE TERMS OF THIS CREATIVE
 * COMMONS PUBLIC LICENSE ("CCPL" OR "LICENSE"). THE WORK IS PROTECTED BY
 * COPYRIGHT AND/OR OTHER APPLICABLE LAW. ANY USE OF THE WORK OTHER THAN AS
 * AUTHORIZED UNDER THIS LICENSE OR COPYRIGHT LAW IS PROHIBITED.
 *
 * BY EXERCISING ANY RIGHTS TO THE WORK PROVIDED HERE, YOU ACCEPT AND AGREE TO
 * BE BOUND BY THE TERMS OF THIS LICENSE. THE LICENSOR GRANTS YOU THE RIGHTS
 * CONTAINED HERE IN CONSIDERATION OF YOUR ACCEPTANCE OF SUCH TERMS AND
 * CONDITIONS.
 *
 */
```

and got

```
src/slam_gmapping.cpp: copyright=Willow Garage, Inc. (2008), Open Source Robotics Foundation, Inc. (2017), license=<unknown>
```

which seems to be correct (since we don't have a BSD template).

I do not know enough to call this resolved, however.

connects to #75

---
This change is [Reviewable](https://reviewable.io/reviews/ament/ament_lint/82)

### DIFF
--- a/ament_copyright/ament_copyright/parser.py
+++ b/ament_copyright/ament_copyright/parser.py
@@ -175,7 +175,7 @@ def search_copyright_information(content):
     year = '\d{4}'
     year_range = '%s-%s' % (year, year)
     year_or_year_range = '(?:%s|%s)' % (year, year_range)
-    pattern = '^[^\n\r]?\s*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
+    pattern = '^[^\n\r]?[\s\*\w]*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
         (year_or_year_range, year_or_year_range)
     regex = re.compile(pattern, re.DOTALL | re.MULTILINE)
 
@@ -198,7 +198,8 @@ def scan_past_coding_and_shebang_lines(content):
     while (
         is_comment_line(content, index) and
         (is_coding_line(content, index) or
-         is_shebang_line(content, index))
+         is_shebang_line(content, index) or
+         is_empty_line(content, index))
     ):
         index = get_index_of_next_line(content, index)
     return index
@@ -239,7 +240,7 @@ def is_shebang_line(content, index):
 
 def get_comment_block(content, index):
     # regex for matching the beginning of the first comment
-    pattern = '^(#|//)'
+    pattern = '^(#|//|/\*\*|/\*)'
     regex = re.compile(pattern, re.MULTILINE)
 
     match = regex.search(content, index)
@@ -248,6 +249,8 @@ def get_comment_block(content, index):
     comment_token = match.group(1)
     start_index = match.start(1)
 
+    if comment_token in ['/*', '/**']:
+        comment_token = ' *'
     end_index = start_index
     while True:
         end_index = get_index_of_next_line(content, end_index)
--- a/ament_copyright/ament_copyright/parser.py
+++ b/ament_copyright/ament_copyright/parser.py
@@ -175,7 +175,7 @@ def search_copyright_information(content):
     year = '\d{4}'
     year_range = '%s-%s' % (year, year)
     year_or_year_range = '(?:%s|%s)' % (year, year_range)
-    pattern = '^[^\n\r]?[\s\*\w]*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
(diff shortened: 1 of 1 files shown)


ament_copyright/ament_copyright/parser.py
```diff
@@ -175,7 +175,7 @@ def search_copyright_information(content):
     year = '\d{4}'
     year_range = '%s-%s' % (year, year)
     year_or_year_range = '(?:%s|%s)' % (year, year_range)
-    pattern = '^[^\n\r]?\s*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
+    pattern = '^[^\n\r]?[\s\*\w]*Copyright(?:\s+\(c\))?\s+(%s(?:,\s*%s)*),?\s+([^\n\r]+)$' % \
```
allenh1: I'm very hesitant to allow for arbitrary alpha-numeric characters here... I don't know enough about licensing to state if this is valid.

ament_copyright/ament_copyright/parser.py
```diff
@@ -248,6 +249,8 @@ def get_comment_block(content, index):
     comment_token = match.group(1)
     start_index = match.start(1)
 
+    if comment_token in ['/*', '/**']:
+        comment_token = ' *'
```
allenh1: This is rather unfortunate -- does anyone see a better way?

ament_copyright/ament_copyright/parser.py
dirk-thomas: Why is this necessary? Is this aiming to allow newlines anywhere in the comment? Would it be sufficient to only make it flexible within the license?

allenh1: This is because of things like

```
/*
 * My Amazing Project
 * Copyright (c) 2017
```

I saw a lot of that in some old willow code (like GMapping).

ament_copyright/ament_copyright/parser.py
dirk-thomas: I think the leading space should not be required. This should be valid:

```
/**
* text
*/
```

ament_copyright/ament_copyright/parser.py
allenh1: I don't think the `\*` is needed, however.

ament_copyright/ament_copyright/parser.py
allenh1: Ok -- I can add that functionality.

```
/**
text
*/
```

is that also acceptable?

allenh1: I'm thinking something like this:

```py
if comment_token in ['/*', '/**']:
    comment_token = '*'
"""
more code
"""
lines = [line[line.index(comment_token) + 1:] for line in lines]
```

That works for the existing copyright cases, but not for the new ones.

allenh1:
I've determined I don't have the time to make this as robust as I'd like, as it will require a very significant rewrite in the `parser.py` file (many assumptions are made based on the fact that the comments start with the same prefix on every line, and that they are on every line).

This PR only supports

```
/*
 *
 */
```

because of the ` *` on the end of the last comment line. This does, however, add the desired newline robustness.

mikaelarguedas:
@allenh1 if you are not working on this anymore, can you reflect it in the labels? (by removing the 'in progress' label from this and the parent issue)

allenh1:
@mikaelarguedas done

mikaelarguedas:
@wjwwood FYI:
This works for bloc comments right now.
```
/*
 * stuff
 */

```
__Current cases not covered:__
- more than 1 space between comment marker and text:
```
/*
 *      stuff
 */
```
- no space between comment marker and text:
```
/*
 *stuff
 */
```
- not leading ` *` for each line within the block comment
```
/*
stuff
*/
```
- not leading space before `*`
```
/*
* stuff
*/
```

greimela-si:
@wjwwood @mikaelarguedas Are there any plans to merge this?

wjwwood:
It's not on my todo-list right now. I doubt I'll have time to address it in the short-term.

dirk-thomas:
Closing due to no activity.

Recent related work: #160.

PR #83, opened 2017-08-25
implicitly inherit from object (only in files not copied from somewhe…

mikaelarguedas:
…re else)

connects to ros2/rclpy#106

### DIFF
--- a/ament_lint_cmake/ament_lint_cmake/cmakelint.py
+++ b/ament_lint_cmake/ament_lint_cmake/cmakelint.py
@@ -78,7 +78,7 @@ _ERROR_CATEGORIES = """\
 """
 _DEFAULT_CMAKELINTRC = os.path.join(os.path.expanduser('~'), '.cmakelintrc')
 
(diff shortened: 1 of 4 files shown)


ament_lint_cmake/ament_lint_cmake/cmakelint.py
```diff
@@ -78,7 +78,7 @@
 """
 _DEFAULT_CMAKELINTRC = os.path.join(os.path.expanduser('~'), '.cmakelintrc')
 
-class _CMakeLintState(object):
+class _CMakeLintState:
```
dirk-thomas: This file is an external project. Therefore I don't think we should modify it.

mikaelarguedas: oh I missed that, thanks for catching it https://github.com/ament/ament_lint/pull/83/commits/abefe7dfe39d8ee20e2ff04eb359210e79bc8128

PR #84, opened 2017-08-29
update style to satisfy new flake8 plugins

dirk-thomas:
Connect to ros2/ci#98.

### DIFF
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -124,8 +124,8 @@ def main(argv=sys.argv[1:]):
             parser.error(
                 "'LICENSE' argument must be a known license name. "
                 "Use the '--list-licenses' options to see alist of valid license names.")
-        license = licenses[args.add_missing[1]]
-        add_missing_header(file_descriptors, name, license, args.verbose)
+        add_missing_header(
+            file_descriptors, name, licenses[args.add_missing[1]], args.verbose)
         return 0
 
     if args.add_copyright_year is not None:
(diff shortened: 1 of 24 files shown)


ament_flake8/ament_flake8/configuration/ament_flake8.ini
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = D100,D101,D102,D103,D104,D105,D203,D212,D404
+ignore = C816,D100,D101,D102,D103,D104,D105,D203,D212,D404
```
mikaelarguedas: is this ignored because it conflicts with another enforced rule (C815?) ?

dirk-thomas: No, `C816` is a suggestion specific to Python 3.6+. Since we are targeting only 3.5 we can't use the suggested style.

ament_copyright/ament_copyright/main.py
```diff
@@ -124,8 +124,8 @@ def main(argv=sys.argv[1:]):
             parser.error(
                 "'LICENSE' argument must be a known license name. "
                 "Use the '--list-licenses' options to see alist of valid license names.")
-        license = licenses[args.add_missing[1]]
-        add_missing_header(file_descriptors, name, license, args.verbose)
+        add_missing_header(
+            file_descriptors, name, licenses[args.add_missing[1]], args.verbose)
```
dhood: what was the error we're addressing here? if it's that we don't want 'wasteful' variables I don't think we should enforce that across the whole codebase as a general rule, because it's usually done for improved readability. but it may have been for another reason.

dirk-thomas: `license` is a built-in in Python and should not be used as a variable name.

PR #85, opened 2017-09-28
update uncrustify config

dirk-thomas:
Connect to ament/uncrustify#18.

### DIFF
--- a/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
+++ b/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
@@ -1,1752 +1,2110 @@
-# Uncrustify 0.62
+# Uncrustify_d-0.65-225-ff0808b
 
 #
 # General options
 #
 
-# The type of line endings
-newlines                                  = auto     # auto/lf/crlf/cr
+# The type of line endings. Default=Auto.
+newlines                        = auto     # auto/lf/crlf/cr
 
-# The original size of tabs in the input
-input_tab_size                            = 8        # number
+# The original size of tabs in the input. Default=8.
+input_tab_size                  = 8        # unsigned number
 
-# The size of tabs in the output (only used if align_with_tabs=true)
-output_tab_size                           = 8        # number
+# The size of tabs in the output (only used if align_with_tabs=true). Default=8.
+output_tab_size                 = 8        # unsigned number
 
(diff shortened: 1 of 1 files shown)


ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
```diff
@@ -1,1752 +1,2110 @@
...
+# True: When identing after virtual brace open and newline add further spaces after regular indent to reach next tabstop.
+indent_vbrace_open_on_tabstop   = false    # false/true
+
+# If True, a brace followed by another token (not a newline) will indent all contained lines to match the token.Default=True.
+indent_token_after_brace        = false    # false/true
 
-# TRUE: When identing after virtual brace open and newline add further spaces after regular indent to reach next tabstop.
-indent_vbrace_open_on_tabstop             = false    # false/true
+# If True, cpp lambda body will be indentedDefault=False.
+indent_cpp_lambda_body          = true     # false/true
```
dirk-thomas: This will indent the body of lambdas by one level which makes them fit in much nicer with the surrounding code.

ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
```diff
@@ -1,1752 +1,2110 @@
...
+sp_before_oc_block_caret        = ignore   # ignore/add/remove/force
 
 # Add or remove space after a block pointer caret
-# '^int (int arg){...}' vs. '^ int (int arg){...}'
-sp_after_oc_block_caret                   = ignore   # ignore/add/remove/force
+# '^int (int arg){...}' vs. '^ int (int arg){...}'.
+sp_after_oc_block_caret         = ignore   # ignore/add/remove/force
 
 # Add or remove space between the receiver and selector in a message.
-# '[receiver selector ..
```
dirk-thomas: This new option enforces that if the parameters for a functions don't fit in a single line (so some parameters are wrapped) that *all* parameters are wrapped. This is the style we are already aiming for in Python.

dirk-thomas: This new option enforces that if the parameters for a functions don't fit in a single line (so some parameters are wrapped) that all parameters are wrapped. This is the style we are already aiming for in Python.

dirk-thomas:
The second commit only changes the order of the option groups (which has changed in uncrustify).

PR #86, opened 2017-10-09
Add new D107 linter error to ignored list

dhood:
This is a new error code that has been introduced in pydocstyle v2.1.0: https://github.com/PyCQA/pydocstyle/pull/277 (previously the errors were reported as D102).

CI including turtlebot repos that (mostly) only runs linters: [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/100/)

http://ci.ros2.org/job/ci_turtlebot-demo_linux/100/testReport/(root)/test_flake8/ compared to http://ci.ros2.org/view/nightly/job/nightly_turtlebot-demo_linux_release/90/testReport/(root)/test_flake8/ in the last nightly


dirk-thomas:
Please also add the also new error code D106 to the ignore list since it is unlikely that we want to enforce it since we already ignore all other documentation related codes.

dhood:
sure thing: cc502025ed9631be09d1e298ae844bddf9be1601

PR #87, opened 2017-10-18
add filters argument to cpplint

dirk-thomas:
Necessary for ros2/rosidl#241.

Ready for review.


dirk-thomas:
Thanks for the fix-ups.

PR #90, opened 2017-12-07
Increase cpplint timeout

mikaelarguedas:
cpplint on arm is hitting the 60-second limit for some packages (http://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_release/254/).

This  exposes the timeout argument and sets it to 120 by default so that we're safe in the future. 

Looking at the time of all linters, this seems to be the only one that needs to be bumped. But I can propagate it to the other linters if we want to have the same set of parameters everywhere.

* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/905/)

### DIFF
--- a/ament_cmake_cpplint/cmake/ament_cpplint.cmake
+++ b/ament_cmake_cpplint/cmake/ament_cpplint.cmake
@@ -24,13 +24,15 @@
 # :type MAX_LINE_LENGTH: integer
 # :param ROOT: override the --root option of cpplint
 # :type ROOT: string
+# :param TIMEOUT: the test timeout in seconds, default: 60
+# :type TIMEOUT: integer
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@wjwwood FYI

ament_cmake_cpplint/cmake/ament_cpplint.cmake
```diff
@@ -24,13 +24,15 @@
 # :type MAX_LINE_LENGTH: integer
 # :param ROOT: override the --root option of cpplint
 # :type ROOT: string
+# :param TIMEOUT: the test timeout in seconds, default: 60
```
dirk-thomas: The documented default value doesn't match the implementation.

PR #98, opened 2018-04-02
Add ament package for pclint

jpsamper2009:
@wjwwood @mjcarroll @dejanpan

- As promised, here is a copy of our `ament_pclint` and `ament_cmake_pclint` packages. I haven't made any changes yet (not even the licenses) because I figure it'll have to go through at least one round of review anyway.

- Neither of these will work if you don't have pclint plus installed (you can buy a license at http://www.gimpel.com)

- The configuration files are modified versions of the ones found here: http://www.gimpel.com/html/ptch90.htm


ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+    except subprocess.CalledProcessError as e:
+        print("The invocation of 'pclint' failed with error code %d: %s" %
+              (e.returncode, e), file=sys.stderr)
+        return False
+    result[idx] = xml
+    retvals[idx] += p.returncode
+
+
+def find_executable(file_name, additional_paths=None):
+    if sys.platform == 'linux':
```
jpsamper2009: There's a small hack here because the binaries `pclp64` is the Windows version, while linux and osx are distributed `pclp64_linux` and `pclp64_osx`.

ament_cmake_pclint/package.xml
```diff
@@ -0,0 +1,25 @@
...
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ament_cmake_pclint</name>
+  <version>0.0.3</version>
+  <description>
+    The CMake API for ament_pclint to perform static code analysis on C/C++
+    code using PCLint.
+  </description>
+  <maintainer email="jp.samper@apex.ai">Juan Pablo Samper</maintainer>
+  <license>Apex.AI Proprietary License</license>
```
dirk-thomas: From this patch it is unclear what this license is. I think each package needs at least a `LICENSE` file.

jpsamper2009: @dirk-thomas It will probably become Apache 2.0 (or another well-known license), I just copy-pasted the code for now, so you could start looking at the content.

ament_cmake_pclint/cmake/ament_pclint.cmake
```diff
@@ -0,0 +1,145 @@
...
+    message(FATAL_ERROR "ament_pclint() could not find program 'ament_pclint'")
+  endif()
+
+  set(cmd "${ament_pclint_BIN}" ${ARG_UNPARSED_ARGUMENTS})
+
+  get_property(INCLUDED_DIRS DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+
+  if(INCLUDED_DIRS)
+      list(APPEND cmd "--include-directories" ${INCLUDED_DIRS})
+      set(_INCLUDE_DIRS_FLAG_SET "Found")
```
dirk-thomas: Nitpick: indentation.

Same below.

ament_cmake_pclint/cmake/ament_pclint.cmake
```diff
@@ -0,0 +1,145 @@
...
+function(_parse_pclint_input)
+
+  find_program(ament_pclint_BIN NAMES "ament_pclint")
+  if(NOT ament_pclint_BIN)
+    message(FATAL_ERROR "ament_pclint() could not find program 'ament_pclint'")
+  endif()
+
+  set(cmd "${ament_pclint_BIN}" ${ARG_UNPARSED_ARGUMENTS})
+
+  get_property(INCLUDED_DIRS DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
```
dirk-thomas: How does it handle the case that no global include directories are being used but target specific include directories instead?

Same for compiler definitions below.

ament_cmake_pclint/cmake/ament_pclint.cmake
```diff
@@ -0,0 +1,145 @@
...
+
+  if(ARG_LNT_FILE)
+    get_filename_component(config_dir ${ARG_LNT_FILE} DIRECTORY)
+    list(APPEND cmd "--pclint-config-dir" ${config_dir})
+    get_filename_component(config_file ${ARG_LNT_FILE} NAME)
+    list(APPEND cmd "--pclint-config-file" ${config_file})
+  endif()
+
+  set(cmd ${cmd} PARENT_SCOPE)
+
```
dirk-thomas: Nitpick: remove the empty lines right after `function()` and before `endfunction()`.

ament_cmake_pclint/package.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ament_cmake_pclint</name>
+  <version>0.0.3</version>
```
dirk-thomas: The version should match the other packages in this repo which are currently 0.4.0.

Same in the other package as well as the `setup.py` file.

ament_pclint/package.xml
```diff
@@ -0,0 +1,18 @@
...
+  <name>ament_pclint</name>
+  <version>0.0.3</version>
+  <description>
+    The ability to perform static code analysis on C/C++ code using pclint
+    and generate xUnit test result files.
+  </description>
+  <maintainer email="jp.samper@apex.ai">Juan Pablo Samper</maintainer>
+  <license>Apex.AI Proprietary License</license>
+
+  <exec_depend>pclp64</exec_depend>
```
dirk-thomas: For the release this needs to be a valid `rosdep` key. If that is not feasible the dependency needs to be removed.

ament_pclint/doc/index.rst
```diff
@@ -0,0 +1,20 @@
...
+.. code:: sh
+
+    ament_pclint [<path> ...]
+
+
+How to run the check from within a CMake ament package as part of the tests?
+----------------------------------------------------------------------------
+
+The CMake integration is provided by the package `ament_cmake_pclint
+<>`_.
```
dirk-thomas: URL of the link?

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+                xml += """  <testcase
+    name=%(quoted_name)s
+    classname="%(testname)s"
+  >
+      <failure message=%(quoted_message)s/>
+  </testcase>
+""" % data
+
+        else:
+            # if there are no cpplint errors report a single successful test
```
dirk-thomas: Copy-n-paste: `cpplint`

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+        print("The invocation of 'pclint' failed with error code %d: %s" %
+              (e.returncode, e), file=sys.stderr)
+        return False
+    result[idx] = xml
+    retvals[idx] += p.returncode
+
+
+def find_executable(file_name, additional_paths=None):
+    if sys.platform == 'linux':
+        file_name = file_name + '_linux'
```
dirk-thomas: Please use `+=` instead.

Same below.

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+    elif sys.platform == 'darwin':
+        file_name = file_name + '_osx'
+    path = None
+    if additional_paths:
+        path = os.getenv('PATH', os.defpath)
+        path += os.path.pathsep + os.path.pathsep.join(additional_paths)
```
dirk-thomas: `additional_paths` doesn't seem to be used anywhere. Should it just be removed? I was mostly wondering if appending of additional paths is the  "right" way...

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+def execute_analysis(cmd, result, retvals, idx):
+    try:
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        xml = p.communicate()[0].decode("utf-8")  # Read xml from stdout
+    except subprocess.CalledProcessError as e:
+        print("The invocation of 'pclint' failed with error code %d: %s" %
+              (e.returncode, e), file=sys.stderr)
+        return False
+    result[idx] = xml
+    retvals[idx] += p.returncode
```
dirk-thomas: Why is this `+=` rather than `=`?

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+            report[filename].append(data)
+
+            data = dict(data)
+            data['filename'] = filename
+            print('[%(filename)s:%(line)d]: (%(severity)s: %(id)s) %(msg)s' % data,
+                  file=sys.stderr)
+
+    # output summary
+    error_count = sum(len(r) for r in report.values())
+    error_count += sum(retvals)
```
dirk-thomas: What is the semantic of each `retval` here? Does it really return the number of errors? That might be problematic when that return code wraps depending on the range or is even negative in case of signals?

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+import subprocess
+import sys
+import time
+from xml.etree import ElementTree
+from xml.sax.saxutils import escape
+from xml.sax.saxutils import quoteattr
+
+
+def main(argv=sys.argv[1:]):
+    extensions = ['c', 'cc', 'cpp', 'cxx', 'c++']
```
dirk-thomas: This list doesn't seem to match the one in the CMake package?

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+    parser.add_argument(
+        'paths',
+        nargs='*',
+        default=[os.curdir],
+        help='Files and/or directories to be checked. Directories are searched recursively for '
+             'files ending in one of %s.' %
+             ', '.join(["'.%s'" % e for e in extensions]))
+    parser.add_argument(
+        '--language',
+        help="Force pclint to analyze files as specified language. Supported languages c or cpp")
```
dirk-thomas: Could this provide a `choices`?

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+    if os.name == 'nt':
+        parser.add_argument(
+            '--pclint-config-dir',
+            nargs=1,
```
dirk-thomas: The default value of `nargs` is 1 so no need to specify it explicitly.

Same below.

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # Prepare pclint command
+    base_cmd = [pclint_bin,
+                '-i"%s"' % args.pclint_config_dir[0],
+                '-i"%s"' % os.path.join(os.path.dirname(os.path.realpath(__file__)), "config"),
+                'env-xml.lnt']
+
+    # Add include directories from arguments
+    if args.include_directories is not None:
+        for directory in args.include_directories:
```
dirk-thomas: Suggestion to avoid the extra conditional:

```
for directory in (args.include_directories or []):
```

Same below.

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # try to determine the number of CPU cores
+    jobs = 1
+    try:
+        jobs = multiprocessing.cpu_count()
+    except NotImplementedError:
+        # the number of cores cannot be determined, don't parallelize
+        pass
+
+    result = ['' for f in files]
```
dirk-thomas: Suggestion: `result = [''] * len(files)`

Same below.

ament_pclint/ament_pclint/main.py
```diff
@@ -0,0 +1,372 @@
...
+             ', '.join(["'.%s'" % e for e in extensions]))
+    parser.add_argument(
+        '--language',
+        help="Force pclint to analyze files as specified language. Supported languages c or cpp")
+    if os.name == 'nt':
+        parser.add_argument(
+            '--pclint-config-dir',
+            nargs=1,
+            type=str,
+            default=[os.path.dirname(os.path.realpath(__file__)) + "/config/msvc"],
```
dirk-thomas: Please use single quotes rather than double quotes where possible without escaping.

Same below.

ament_cmake_pclint/cmake/ament_pclint.cmake
jpsamper2009: I'm not 100% I understand the question, but I think what I meant to do here was gather all the directories included through cmake's `include_directories`. The idea is that you need to tell pclint where to look for header files, not unlike how you have to tell cmake where to find header files. 

It's likely that I'm misusing or misunderstanding something in cmake, though.

jpsamper2009: Similarly, we want to give pclint compiler definitions added through target_compile_definitions, but again, I may be misunderstanding something fundamental about cmake.

dirk-thomas: The CMake code might set these information on a per-targetg level using e.g. `target_include_directories` and/or `target_compile_definitions`. In that case the global variables might not be sufficient.

ament_pclint/package.xml
jpsamper2009: My guess is this won't be the case since it's a proprietary license, so I'll remove.

ament_pclint/ament_pclint/main.py
jpsamper2009: I was just following what `ament_cppcheck` is doing: https://github.com/ament/ament_lint/blob/master/ament_cppcheck/ament_cppcheck/main.py#L158-L163

But, it's up to you, I can delete it.

ament_pclint/ament_pclint/main.py
jpsamper2009: It was probably a single retval as some point, and I forgot to remove the += when I changed it to indeces.

ament_pclint/ament_pclint/main.py
jpsamper2009: Do you mean the list in `ament_cmake_pclint_lint_hook.cmake` (https://github.com/ament/ament_lint/pull/98/files#diff-ebf9131ad30342b5b69552b98477905a)? You are right, it's missing the `.c++` extension.

ament_pclint/ament_pclint/main.py
jpsamper2009: I should probably be using os.path.join, no?

ament_pclint/ament_pclint/main.py
dirk-thomas: I am not saying that it should be removed. I just want to point out that it might not work for packages which use target specific properties (which is the modern way in CMake).

ament_pclint/ament_pclint/main.py
dirk-thomas: Even better :+1:

ament_cmake_pclint/cmake/ament_pclint.cmake
jpsamper2009: Done.

ament_cmake_pclint/cmake/ament_pclint.cmake
jpsamper2009: Done

ament_cmake_pclint/package.xml
jpsamper2009: Done.

ament_pclint/doc/index.rst
jpsamper2009: Done

ament_pclint/ament_pclint/main.py
jpsamper2009: Oops :) fixed

ament_pclint/ament_pclint/main.py
jpsamper2009: Done

ament_pclint/ament_pclint/main.py
jpsamper2009: Done

ament_pclint/ament_pclint/main.py
jpsamper2009: Done

ament_pclint/ament_pclint/main.py
jpsamper2009: Done

ament_pclint/ament_pclint/main.py
jpsamper2009: Done

jpsamper2009:
@dirk-thomas Do you have a suggestion on how to gracefully skip this test if pclint is not available? I'm thinking I can either check at the cmake level if the executable exists, and if it doesn't print a Warning and continue. I could also just do it at the ament_pclint.py level where if no executable is found, I print a warning and exit with return code 0.

dirk-thomas:
I would propose to set the `DISABLED` property on the CMake test when `pclint` was not found without printing a warning. Otherwise when the linter is added to the defaults every user would see a warning for every package when `pclint` is not available. While this new property is only available as of [CMake 3.9](https://cmake.org/cmake/help/v3.9/prop_test/DISABLED.html) that should be fine since the next release can bump the minimum required version of CMake since Bionic ships with 3.10.

For the command line tool I would rather return a non-zero error code when `pclint` is unavailable. Passing without actually running the linter doesn't sounds like a good idea to me.

ament_cmake_pclint/cmake/ament_pclint.cmake
jpsamper2009: I see what you're saying now. I was assuming that any headers are included with `include_directories`

What I'm thinking is we need to add PCLINT_INCLUDE_DIRS and PCLINT_COMPILE_DEFS options to `ament_pclint` or do you know of something better to get information added via `target_*()`? 

Basically, pclint needs access to the exact same directories as the `add_executable` or `add_library` functions.

dirk-thomas: I don't see a better way then to let the user pass in the necessary flags manually. That can be either the include dirs directory or a list of targets from which the include dirs are being extracted.

Beside supporting these information as arguments in the CMake function the user also needs to be able to set some CMake variables for the case where the linter is being invoked implicitly by the `ament_lint_auto` package.

mjcarroll: I was considering how to do this as well.  @jpsamper2009 is there not a way to retrieve all of the includes and flags for a particular target rather than the global scope in CMake?

mjcarroll:
@jpsamper2009 I have been using it. My biggest things were the exit without the executable being found, and the handling of the compiler flags.

I was also looking into retrieving the target flags for the open-source `cppcheck` tool for the open-source CI side.

ament_cmake_pclint/cmake/ament_pclint.cmake
mjcarroll: I think what you want to do is:

add_library(
  ${PROJECT_NAME}
  ${rcutils_sources})

```
get_target_property(TARGET_INC ${PROJECT_NAME} INCLUDE_DIRECTORIES)
message(STATUS "INCLUDES: " ${TARGET_INC})

get_target_property(TARGET_FLAGS ${PROJECT_NAME} COMPILE_FLAGS)
message(STATUS "FLAGS: " ${TARGET_FLAGS})

get_target_property(TARGET_DEF ${PROJECT_NAME} COMPILE_DEFINITIONS)
message(STATUS "DEFINITIONS: " ${TARGET_DEF})
```

And then use those generated variables to pass into pclint.

mjcarroll: Upon further reflection, I think that the appropriate way to handle `ament_cmake_pclint` is to add a CMake function that takes a target as an argument.  From there, we can retrieve the list of sources, includes, and compiler definitions/flags that are used on that target, to make the linting maximally effective.

This would also allow us to exclude things like unit tests, etc from pclint checking.

jpsamper2009:
@mjcarroll @dirk-thomas I've reworked this a bit to get rid of some unused parts and address your suggestions. I'm afraid I didn't see https://github.com/ament/ament_lint/pull/98#discussion_r181906107 until just now, but maybe the current solution is sufficient to merge this, and we can work on reading information from a target on a separate issue? It would be good to get some people using it also to see what use cases we actually have instead of implementing a bunch of features that then go unused.

ament_cmake_pclint/cmake/ament_pclint.cmake
jpsamper2009: @mjcarroll I hadn't seen your comment above, so for now I've just added function arguments to allow the user to specify the include directories explicitly.

Do you guys have an example of a package with target-specific include directories and compiler flags?

mjcarroll:
What 3.9 features is this using?

jpsamper2009:
@mjcarroll The second commit is using the DISABLE property which is only available in 3.9 (https://github.com/ament/ament_lint/pull/98#issuecomment-378674366). As long as you have pclint installed, you should be able to use the first commit.

@dirk-thomas Ack

mjcarroll:
This has been moved back to in progress.  Because we are going to try to keep some Xenial compatibility, we need to find a way to accomplish the `DISABLE` property without using cmake 3.9

dirk-thomas:
(or just skip creating the test)

jpsamper2009:
@mjcarroll @dirk-thomas I have added a check to skip the test creation and print a warning if the executable is not found: (https://github.com/ament/ament_lint/pull/98/commits/c3eaaf484561b7dd16c556b165fad65cf56cf091)

I have also put the CMake 3.9 solution into a different branch in case we want to go back to that version eventually: https://github.com/jpsamper2009/ament_lint/commit/e2936447aa0dababd37605d4f0fa752953cc8799

ament_pclint/setup.py
```diff
@@ -0,0 +1,53 @@
...
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Apex AI',
+        'Programming Language :: Python',
+        'Topic :: Software Development',
+    ],
+    description='Static code analysis on C/C++ code using PCLint.',
+    long_description="""\
+The ability to perform static code analysis on C/C++ code using PCLint
+and generate xUnit test result files.""",
+    license='Apex AI',
```
dirk-thomas: This information doesn't match the package manifest which declares Apache 2.

jpsamper2009:
I've just pushed a change to one of the pclint config files. The file is available online ( http://www.gimpel.com/html/pub90/au-misra3.lnt), but the version we were using had a less permissive license. I've instead adjusted the configuration linked above.

mjcarroll:
@dirk-thomas Sorry about that, I must have misclicked.

serge-nikulin:
@jpsamper2009, 

FYI (unless you guys already fixed it)

I finally found the option to check local headers we missed back then: `+libclass(angle)`. See Chapter 5 in the PCLint manual for details. Add this option to both `c++.lnt` and `c99.lnt` files.

Also, I use a forced check now for all headers even if they are not used by local sources (via creating throw-away trivial users). See below the relevant fragment from my `main.py` file.

```python
# For each header file create a temp .cpp file that includes it via ""
# This way PCLint wil check all header files even for pure interface libraries
def prepare_source_files(source_files, include_files):
    tempdir = tempfile.TemporaryDirectory(prefix='ament_pclintsa_')
    for header_name in include_files:
        if header_name.startswith('include/'):
            # A hack which must be fixed
            # remove "include/" prefix
            header_name = header_name[8:]
        header_name = header_name.replace('\\', '/')
        src_name = os.path.join(tempdir.name, os.path.basename(header_name) + '.cpp')
        with open(src_name, 'w') as f:
            f.write('#include "' + header_name + '"\n')
        source_files.append(src_name)
    # return tempdir too so it will be deleted on the script's end
    return source_files, tempdir
```

dirk-thomas:
@serge-nikulin This ticket has been closed for almost a year. The chances your comment will get lost is very high. Please either create a pull request with your proposed fragment or open an issue with the information (with a strong preference for the former).

PR #103, opened 2018-07-09
update configuration to 0.67

mikaelarguedas:
Connects to ament/uncrustify#24

### DIFF
--- a/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
+++ b/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
@@ -1,4 +1,4 @@
-# Uncrustify-0.66.1-28-788c5ce
+# Uncrustify-0.67
 
 #
 # General options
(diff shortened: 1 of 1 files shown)


ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
```diff
@@ -370,6 +383,9 @@ sp_square_fparen                = remove   # ignore/add/remove/force
 # Add or remove space between ')' and '{' of function.
 sp_fparen_brace                 = force    # ignore/add/remove/force
 
+# Add or remove space between ')' and '{' of function call in object initialization. Overrides sp_fparen_brace.
+sp_fparen_brace_initializer     = ignore   # ignore/add/remove/force
```
dirk-thomas: Since this overrides `sp_fparen_brace` should it be the same value - `force`?

mikaelarguedas: Good point, done in https://github.com/ament/ament_lint/pull/103/commits/43a0307c3f26a8f313e7dae8d69c3d44c84b6bf6

PR #104, opened 2018-07-11
add ament_(cmake_)xmllint packages

dirk-thomas:
https://ci.ros2.org/job/ci_linux/4876/testReport/ as an example of failed linting.

If `xmllint` is not available (as it is the case on the non-Linux Jenkins nodes) the tests are being skipped: https://ci.ros2.org/job/ci_linux/4875/testReport/junit/(root)/projectroot/

With other PRs merged this passes CI: [Build Status](https://ci.ros2.org/job/ci_linux/4882/)

### DIFF
--- a/ament_cmake_xmllint/cmake/ament_xmllint.cmake
+++ b/ament_cmake_xmllint/cmake/ament_xmllint.cmake
@@ -1,0 +1,69 @@
+# Copyright 2014-2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# Add a test to check XML files for validity with xmllint.
+#
+# :param TESTNAME: the name of the test, default: "xmllint"
+# :type TESTNAME: string
+# :param ARGN: the files or directories to check
+# :type ARGN: list of strings
+#
+# @public
+#
+function(ament_xmllint)
+  cmake_parse_arguments(ARG "" "MAX_LINE_LENGTH;TESTNAME" "" ${ARGN})
+  if(NOT ARG_TESTNAME)
+    set(ARG_TESTNAME "xmllint")
+  endif()
+
+  find_program(ament_xmllint_BIN NAMES "ament_xmllint")
+  if(NOT ament_xmllint_BIN)
+    message(FATAL_ERROR "ament_xmllint() could not find program 'ament_xmllint'")
+  endif()
+
+  set(result_file "${AMENT_TEST_RESULTS_DIR}/${PROJECT_NAME}/${ARG_TESTNAME}.xunit.xml")
+  set(cmd "${ament_xmllint_BIN}" "--xunit-file" "${result_file}")
+  list(APPEND cmd ${ARG_UNPARSED_ARGUMENTS})
+
+  find_program(xmllint_BIN NAMES "xmllint")
+
+  if(NOT xmllint_BIN)
+    if(${CMAKE_VERSION} VERSION_LESS "3.8.0")
+      message(WARNING "WARNING: 'xmllint' not found, skipping xmllint test creation")
+      return()
+    endif()
+  endif()
+
+  file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/ament_xmllint")
+  ament_add_test(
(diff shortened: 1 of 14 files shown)


mikaelarguedas:
Does this check the validity of the xml against the schema ? or just xml validity in general (e.g. mismatching/missing tags)?

ament_cmake_xmllint/cmake/ament_xmllint.cmake
```diff
@@ -0,0 +1,69 @@
...
+  endif()
+
+  set(result_file "${AMENT_TEST_RESULTS_DIR}/${PROJECT_NAME}/${ARG_TESTNAME}.xunit.xml")
+  set(cmd "${ament_xmllint_BIN}" "--xunit-file" "${result_file}")
+  list(APPEND cmd ${ARG_UNPARSED_ARGUMENTS})
+
+  find_program(xmllint_BIN NAMES "xmllint")
+
+  if(NOT xmllint_BIN)
+    if(${CMAKE_VERSION} VERSION_LESS "3.8.0")
```
mikaelarguedas: ~~what is the cmake 3.8 feature being used here ?~~ From @dirk-thomas `DISABLED` used L66 is the answer

mikaelarguedas:
```
error : Unknown IO error
warning: failed to load external entity "http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd"
Schemas parser error : Failed to locate the main schema resource at 'http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd'.
WXS schema http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd failed to compile
```

Arg that doesn't look good. I suggest to revert / remove the last commit and merge this as is. A follow-up ticket / PR can be opened with a reference to the commit so that we can look into reenabling it when we find time to look into it.

dirk-thomas:
The url `http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd` seems to redirect to `https`. `xmllint` might not be able to follow that. Please try it with the non-redirected URL.

dirk-thomas:
Can you try it with `xmllint` directly.

dirk-thomas:
With the downloaded schema it works for me: `xmllint ./src/ros2/system_tests/test_security/test/test_security_files/governance.xml --schema omg_shared_ca_governance.xsd`

mikaelarguedas:
```
$ xmllint -schema https://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd test/test_security_files/governance.xml
```

Fails: with
```
Schemas parser error : Failed to locate the main schema resource at 'https://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd'.
WXS schema https://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd failed to compile
```

But this works:
```
$ wget https://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd
$ xmllint -schema ./omg_shared_ca_governance.xsd test/test_security_files/governance.xml 
```

I observe the same behavior with their new schema: https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_governance.xsd

dirk-thomas:
The problem is actually not the redirect. `xmllint` doesn't support `https` URLs. See e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=679867

Do you want to find the same schema on a http url and use that, or host the schema on http://download.ros.org/schema/ or should we disable the linter for that package?

mikaelarguedas:
Let's disable that package for now as it would fail the test even if it was finding the schema properly.

We can revisit later, either use a linter that supports https or a non-secured hosting solution for these xsds

dirk-thomas:
> Let's disable that package for now as it would fail the test even if it was finding the schema properly.

Done in ros2/system_tests#287.

dirk-thomas:
Follow up commit 0ad4e0fc0c1e00afed0a3754eee7bbc83c038cea to avoid stderr output.

mikaelarguedas:
@dirk-thomas These linter tests ran on MacOS last night (I was not expecting it based on the PR description).
Do we know what system dependency needs to be installed to get the xmllint binary ?

dirk-thomas:
> Do we know what system dependency needs to be installed to get the xmllint binary ?

`xmllint` - on Ubuntu it is part of the Debian package `libxml2-utils` (see ros2/ci#207).

mikaelarguedas:
Does this mean that the package `xmllint` has been installed from `brew` on all MacOS nodes? (if so can you please update the buildfarmer logbook)
But not on Windows because no choco packages for it?

I'm asking because we will need this information to update the node setup instructions and eventually user install instructions

dirk-thomas:
> Does this mean that the package xmllint has been installed from brew on all MacOS nodes? (if so can you please update the buildfarmer logbook)

I have not touched any macOS node. It seems the executable was already available.

> But not on Windows because no choco packages for it?

I didn't see a choco package but didn't look for long since it should be sufficient if one platform runs the linter.

PR #106, opened 2018-08-21
Bsd clause3 fixup

mikaelarguedas:
As discussed offline, we need to relax the matching of the 3rd clause of the BSD 2 license to make porting easier.
This PR removes the requirement for the `the` preceding the copyright holder.

The second commit is cosmetic and replace the `{company}` placeholder with `{copyright_holder}` as the copyright holder doesn't have to be a company

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5080/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1867/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4213/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5064/)

### DIFF
--- a/ament_copyright/ament_copyright/parser.py
+++ b/ament_copyright/ament_copyright/parser.py
@@ -62,7 +62,7 @@ class FileDescriptor:
         for name, license_ in get_licenses().items():
             template = getattr(license_, license_part).replace('\n', ' ').strip()
             last_index = -1
-            for license_section in template.split('{company}'):
+            for license_section in template.split('{copyright_holder}'):
                 # OK, now look for each section of the license in the incoming
                 # content.
                 index = content.replace('\n', ' ').strip().find(license_section.strip())
--- a/ament_copyright/ament_copyright/template/bsd2_header.txt
+++ b/ament_copyright/ament_copyright/template/bsd2_header.txt
@@ -13,7 +13,7 @@ are met:
    copyright notice, this list of conditions and the following
    disclaimer in the documentation and/or other materials provided
    with the distribution.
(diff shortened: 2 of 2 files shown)


dirk-thomas:
Maybe using a "generic" license text as in https://opensource.org/licenses/BSD-3-Clause should be considered for the template. It also insulates the expansion of the copyright holder in the leading lines to affect the license text below.

mikaelarguedas:
Yes that was my original proposal. The reason it was not implemented is because the ported ROS 1 code shouldn't have to change the content of the license and currently all ROS 1 code lists Willow Garage Inc. in the third clause instead of the generic "copyright holder" term.

That is why we decided to accept basically anything between  `Neither the name of` and `nor the names of its` to account for all the code out there that is expanding the copyright holder in the 3rd clause

dirk-thomas:
> Yes that was my original proposal.

I would second that.

> The reason it was not implemented is because the ported ROS 1 code shouldn't have to change the content of the license

I am not sure that is a strict requirement. If slightly modified licence blocks need to be "unified" during the process that would be acceptable (even preferable) to me. (I would prefer a strict comparison over the "alow anything in between - I could simply write a very different license block which would still be classifies as BSD...)

dirk-thomas:
[I am not objecting to merge this change - I just don't think we have to make the template as flexible as possible to allow "anything" similar to pass. At the end the developer opt-in to use the copyright linter.]

mikaelarguedas:
Opened #107 to track alternatives. Merging this

PR #110, opened 2018-10-24
Ignore flake8 W504

jacobperron:
Resolves ros2/build_cop#149.

### DIFF
--- a/ament_flake8/ament_flake8/configuration/ament_flake8.ini
+++ b/ament_flake8/ament_flake8/configuration/ament_flake8.ini
@@ -1,5 +1,5 @@
 [flake8]
-ignore = C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
+ignore = C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202,W504
 import-order-style = google
 max-line-length = 99
 show-source = true


jacobperron:
Specifically, this ignores **"line break after binary operator"**.

jacobperron:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5339/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2074/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4437/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5307/)

jacobperron:
Looks like **W605 "invalid escape sequence"** is affecting us. It was also introduced in `flake8 3.6` (using [pycodestyle 2.4(https://github.com/PyCQA/pycodestyle/blob/c3d2cbd744236c3a41d1013c9dce2712dcc4eee0/CHANGES.txt#L18)). Seems like most, of the warnings are due to regex's and can be resolved by using raw strings. We can go forward and fix the warnings in the affected packages or add another ignore here.

It does not appear that W606 or E252 (also new in pycodestyle 2.4) are causing any issues.

dirk-thomas:
If it is only in a few cases I would rather lean towards addressing them.

PR #117, opened 2018-12-18
Pass include paths to cppcheck

### LINKED ISSUE #116: [ament_cppcheck] Unable to resolve macro definitions from includes
With the latest release of cppcheck (v1.86), tests have started failing in `rclcpp` with the following error message (or very similar):

```
There is an unknown macro here somewhere. Configuration is required. If RCLCPP_SMART_PTR_DEFINITIONS is a macro then please configure it.
```

Example build: https://ci.ros2.org/view/All/job/test_ci_osx/118/

---

I don't believe this is an error in `rclcpp` itself, but with how cppcheck is invoked.
If cppcheck is invoked with the `-I` option, passing the include directory, then the error goes away.
E.g.
```
cppcheck -I rclcpp/rclcpp/include  rclcpp/rclcpp/include/rclcpp/client.hpp
```

Documentation for the `-I` option:

```
    -I <dir>             Give path to search for include files. Give several -I
                         parameters to give several paths. First given path is
                         searched for contained header files first. If paths are
                         relative to source files, this is not needed.
```

jacobperron:
Specifically, passing the parent directories of all source files since a typical include
for a C/C++ header are located in a subdirectory named after the package or library.

---

Resolves #116 

Test CI with cppcheck v1.86 (up to rclcpp): [Build Status](https://ci.ros2.org/view/All/job/test_ci_osx/119/)


ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -86,6 +91,9 @@ def main(argv=sys.argv[1:]):
            '--xml-version=2']
     if args.language:
         cmd.extend(['--language={0}'.format(args.language)])
+    if len(args.include_dirs) > 0:
```
tfoote:
```suggestion
    if args.include_dirs:
```

An empty argument will default to None for which len(None) is undefined. 

```
>>> parser=argparse.ArgumentParser()
>>> parser.add_argument('--asdf', nargs='*')
_StoreAction(option_strings=['--asdf'], dest='asdf', nargs='*', const=None, default=None, type=None, choices=None, help=None, metavar=None)
>>> args = parser.parse_args()
>>> args
Namespace(asdf=None)
>>> len(args.asdf)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()

```

jacobperron: Thanks! Nice catch

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +24,14 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
 )
 if(_source_files)
   message(STATUS "Added test 'cppcheck' to perform static code analysis on C / C++ code")
-  ament_cppcheck()
+  # Get include paths for sources
+  set(_include_dirs "")
+  foreach(_src ${_source_files})
+    get_filename_component(_src_dir ${_src} DIRECTORY)
+    # Get parent directory since includes typically have the pattern 'package/source.hpp'
+    get_filename_component(_parent_dir ${_src_dir} DIRECTORY)
```
dirk-thomas: This heuristic looks rather fragile to me. I can imagine cases where we have multiple levels of nesting where this might even lead to false results.

jacobperron: I discovered an alternative command to this heuristic. Hopefully less fragile. 
aa40c90

dirk-thomas: Setting global include directories is discouraged in favor of target specific properties. So this approach is unlikely going to work on the future.

jacobperron: I'm not sure how else to infer the include directories here, (until cmake 3.7 when we can use [BUILDSYSTEM_TARGETS](https://cmake.org/cmake/help/v3.13/prop_dir/BUILDSYSTEM_TARGETS.html) to get `INCLUDE_DIRECTORIES` for each target).

Alternatively, affected packages can call the ament cppcheck macro explicitly, passing the appropriate include paths.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5943/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2470/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4912/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5835/)

And CI machine with new version of cppcheck (v1.86)
* macOS [Build Status](https://ci.ros2.org/job/test_ci_osx/124/)

jacobperron:
With lint fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5946/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4915/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5838/)

cppcheck v1.86:
* macOS [Build Status](https://ci.ros2.org/job/test_ci_osx/125/)

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
dirk-thomas: > until cmake 3.7 when we can use `BUILDSYSTEM_TARGETS` to get `INCLUDE_DIRECTORIES` for each target

That sounds like the best path forward as soon as we decide to drop Xenial support.

jacobperron:
I've updated the PR to use `BUILDSYSTEM_TARGETS` if cmake version is >= 3.7. We're not experiencing cppcheck issues for systems with older cmake versions, like Xenial, so not passing include directories (as before) should remain okay.
@dirk-thomas Please take another look.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +24,19 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
 )
 if(_source_files)
   message(STATUS "Added test 'cppcheck' to perform static code analysis on C / C++ code")
-  ament_cppcheck()
+
+  # Get include paths for added targets
+  set(_all_include_dirs "")
+  # BUILDSYSTEM_TARGETS only supported in CMake >= 3.7
+  if(CMAKE_VERSION VERSION_EQUAL "3.7.0" OR CMAKE_VERSION VERSION_GREATER "3.7.0")
```
dirk-thomas: Instead of two conditions repeating the version please use only one condition: `NOT CMAKE_VERSION VERSION_LESS "3.7.0"`

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +24,19 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+    foreach(target ${BUILDSYSTEM_TARGETS})
+      get_property(_include_dirs
+        TARGET ${target}
+        PROPERTY INCLUDE_DIRECTORIES
+      )
+      list(APPEND _all_include_dirs "${_include_dirs}")
```
dirk-thomas: Since `_include_dirs` contains a list I don't think it should be quoted.

Same below.

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -86,6 +91,9 @@ def main(argv=sys.argv[1:]):
            '--xml-version=2']
     if args.language:
         cmd.extend(['--language={0}'.format(args.language)])
+    if args.include_dirs:
+        for include_dir in args.include_dirs:
```
dirk-thomas: These two lines can be simplified to: `for include_dir in (args.include_dirs or []):`

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -86,6 +91,9 @@ def main(argv=sys.argv[1:]):
...
+            cmd.extend(['-I', '{0}'.format(include_dir)])
```
dirk-thomas: Why use a format string here when `include_dir` already contains the desired string?

ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
```diff
@@ -38,6 +40,9 @@ function(ament_cppcheck)
   set(result_file "${AMENT_TEST_RESULTS_DIR}/${PROJECT_NAME}/${ARG_TESTNAME}.xunit.xml")
   set(cmd "${ament_cppcheck_BIN}" "--xunit-file" "${result_file}")
   list(APPEND cmd ${ARG_UNPARSED_ARGUMENTS})
+  if(ARG_INCLUDE_DIRS)
+    list(APPEND cmd "--include_dirs" "${ARG_INCLUDE_DIRS}")
```
dirk-thomas: You might want to try this with multiple include dirs which contain spaces in the path.

ament_cppcheck/ament_cppcheck/main.py
jacobperron: Left over from a previous iteration. I'll change it.

ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
jacobperron: Appears to work.

jacobperron:
A couple issues I've identified with this approach:

1. Passing include directories to cppcheck significantly increases the time it takes to run.
It doesn't seem like 120s is enough in some situations. Likely need to bump the timeout >200s (maybe 300s if we can live with it).
2. Errors are reported from included headers from outside the package being tested.
I have a hack to handle this in e67e564.

Thoughts?

jacobperron:
Attempt to resolve timeout issue and errors from external packages by only including headers from the package being tested.

[Build Status](https://ci.ros2.org/job/test_ci_osx/126/)

jacobperron:
Ready for another round of review.

Passing include directories to cppcheck uncovered a potential issue in rcl (fixed in ros2/rcl#371). It's possible there will be more uncovered issues.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5998/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5874/)

jacobperron:
CI is good. The one cppcheck error is handled by ros2/rcl#371.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +24,31 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+      #     1. Reduces execution time (less include directories to search)
+      #     2. cppcheck will not check for errors in external packages
+      foreach(_include_dir ${_include_dirs})
+        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_match ${_include_dir})
+        if(_is_match)
+          list(APPEND _all_include_dirs ${_include_dir})
+        endif()
+      endforeach()
+    endforeach()
+    list(REMOVE_DUPLICATES _all_include_dirs)
```
dirk-thomas: If I remember correctly I looked into this function in the past and at that time it didn't guarantee stability of the result. Since it is important that the order of the include dirs is maintained I would suggest to use [list_append_unique](https://github.com/ament/ament_cmake/blob/2d65d41e047e203cd0531df9b0987d8aaea7d6ca/ament_cmake_core/cmake/core/list_append_unique.cmake#L28) instead of this line and `list(APPEND ...)` above.

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -117,7 +122,10 @@ def main(argv=sys.argv[1:]):
             'severity': error.get('severity'),
             'msg': error.get('verbose'),
         }
-        report[filename].append(data)
+
+        file_report = report.get(filename, [])
+        file_report.append(data)
+        report.setdefault(filename, file_report)
```
dirk-thomas: Instead of calling `get` and `setdefault` this can be simplified to:

```
file_report = report.setdefault(filename, [])
file_report.append(data)
```

As an alternative the type of `report` could be changed to `defaultdict(list)`. That would avoid the extract call altogether.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
jacobperron: I don't think the order of the include directories matters in this case, does it?
I can make the suggested change nonetheless.

dirk-thomas: The order of the include directories will determine which header file will be used if the same file exists in multiple locations.

jacobperron: Done

ament_cppcheck/ament_cppcheck/main.py
jacobperron: Done.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +24,30 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+      )
+
+      # Only append include directories that are from the package being tested
+      # This accomplishes two things:
+      #     1. Reduces execution time (less include directories to search)
+      #     2. cppcheck will not check for errors in external packages
+      foreach(_include_dir ${_include_dirs})
+        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_match ${_include_dir})
+        if(_is_match)
+          list_append_unique(_all_include_dirs ${_include_dir})
```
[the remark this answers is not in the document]
jacobperron: Done ba7108b

jacobperron:
CI (linter tests only):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6006/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2513/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4957/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5881/)

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +26,30 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+  # BUILDSYSTEM_TARGETS only supported in CMake >= 3.7
+  if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
+    get_directory_property(_build_targets DIRECTORY ${CMAKE_SOURCE_DIR} BUILDSYSTEM_TARGETS)
+    foreach(_target ${_build_targets})
+      get_property(_include_dirs
+        TARGET ${_target}
+        PROPERTY INCLUDE_DIRECTORIES
+      )
+
+      # Only append include directories that are from the package being tested
```
sloretz: I'm seeing a similar cppcheck complaint in `rmw_connext` and `rmw_opensplice` where they use a macro `RMW_CHECK_TYPE_IDENTIFIERS_MATCH` that is defined `rmw`.

https://ci.ros2.org/view/nightly/job/nightly_osx_extra_rmw_release/313/testReport/junit/rmw_opensplice_cpp/cppcheck/error__unknownMacro__src_rmw_wait_cpp_80_/

https://ci.ros2.org/job/nightly_osx_debug/1175/testReport/junit/rmw_connext_shared_cpp/cppcheck/error__unknownMacro___Users_osrf_jenkins_agent_workspace_nightly_osx_debug_ws_src_ros2_rmw_connext_rmw_connext_shared_cpp_include_rmw_connext_shared_cpp_wait_hpp_51_/

Any way to tell cppcheck where to look for macros without checking external files?

jacobperron: @sloretz See #125.

You can pass cppcheck one or more include directories that contain the macros it is trying to resolve. 
I'm not satisfied with the solution, but I think anything better would require changes in `cppcheck` itself.

sloretz: thanks, I missed that PR

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -105,9 +113,7 @@ def main(argv=sys.argv[1:]):
         return 1
 
     # output errors
-    report = {}
-    for filename in files:
-        report[filename] = []
+    report = defaultdict(list)
```
dirk-thomas: This change broke the report since only files with error appear in the report now. See #134 for reverting this part.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -24,5 +26,30 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+      # This accomplishes two things:
+      #     1. Reduces execution time (less include directories to search)
+      #     2. cppcheck will not check for errors in external packages
```
dirk-thomas, line 44: This fails on Windows in this case: ros2/ros2#942.

jacobperron: I don't understand the case that's failing. From the referenced ticket, it looks like the failures are due to headers from external packages not being passed to cppcheck (which is the intended behavior). For this case, we have `ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS` that can be set (introduced in #125).

dirk-thomas: Without the proper include directories `cppcheck` warns for unknown macros.

ros2/rclpy#577 sets an additional include dir.

PR #118, opened 2018-12-19
[ament_uncrustify] Account for zero value in line length cmake arg

jacobperron:
A zero value will cause uncrustify to ignore line length.

Connects to ros2/rosidl#340
Connects to ament/uncrustify_vendor#3


dirk-thomas:
I think the config file needs to be updated for the new version. Currently the first line states that it is for version 0.67.

jacobperron:
New uncrustify options were added with defaults. No options were removed (see [uncrustify changelog](https://github.com/uncrustify/uncrustify/blob/master/ChangeLog)).

PR #121, opened 2019-01-11
Account for INTERFACE libraries when getting target include directories

### LINKED ISSUE #120: The   property "INCLUDE_DIRECTORIES" is not allowed.
When trying to compile rosbag2 with the latest ros2 master, I run into the following cmake error. I believe it's related to https://github.com/ament/ament_lint/pull/117

Do we have to change something in the rosbag2 code base?

```
--- stderr: rosbag2_test_common
CMake Error at /root/ros2_ws/install/share/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake:36 (get_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "INCLUDE_DIRECTORIES" is not allowed.
Call Stack (most recent call first):
  /root/ros2_ws/install/share/ament_cmake_core/cmake/core/ament_execute_extensions.cmake:38 (include)
  /root/ros2_ws/install/share/ament_lint_auto/cmake/ament_lint_auto_package_hook.cmake:15 (ament_execute_extensions)
  /root/ros2_ws/install/share/ament_cmake_core/cmake/core/ament_execute_extensions.cmake:38 (include)
  /root/ros2_ws/install/share/ament_cmake_core/cmake/core/ament_package.cmake:66 (ament_execute_extensions)
  CMakeLists.txt:43 (ament_package)
```

The rosbag2 code in question is relatively simple:
https://github.com/ros2/rosbag2/blob/master/rosbag2_test_common/CMakeLists.txt#L20-L41

```
  target_include_directories(${PROJECT_NAME} INTERFACE
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
    $<INSTALL_INTERFACE:include>)
```

When resetting `ament_cmake_cppcheck` to version 0.6.2 it compiles correctly.

jacobperron:
Fixes #120

CMake does not allow getting the INCLUDE_DIRECTORIES property from
INTERFACE libraries.
Instead, first check if the property exists, if it does not then try to
get the INTERFACE_INCLUDE_DIRECTORIES property.

Note, if INTERFACE_INCLUDE_DIRECTORIES is not defined an empty list is
returned, but we cannot assume the target is not an interface.
This is why the implementation is conditional on INCLUDE_DIRECTORIES
instead.

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
@@ -33,24 +33,12 @@ if(_source_files)
   if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
     get_directory_property(_build_targets DIRECTORY ${CMAKE_SOURCE_DIR} BUILDSYSTEM_TARGETS)
     foreach(_target ${_build_targets})
-      # Check if INCLUDE_DIRECTORIES is defined first in case the target is an INTERFACE
-      get_property(_include_dirs_defined
-        TARGET ${_target}
-        PROPERTY INCLUDE_DIRECTORIES
-        DEFINED
-      )
-
-      if(_include_dirs_defined)
-        get_property(_include_dirs
-          TARGET ${_target}
-          PROPERTY INCLUDE_DIRECTORIES
-        )
+      # Include directories property is different for INTERFACE libraries
+      get_target_property(_target_type ${_target} TYPE)
+      if(${_target_type} STREQUAL "INTERFACE_LIBRARY")
+        get_target_property(_include_dirs ${_target} INTERFACE_INCLUDE_DIRECTORIES)
       else()
-        # Target might be an interface
-        get_property(_include_dirs
-          TARGET ${_target}
-          PROPERTY INTERFACE_INCLUDE_DIRECTORIES
-        )
+        get_target_property(_include_dirs ${_target} INCLUDE_DIRECTORIES)
       endif()
 
       # Only append include directories that are from the package being tested
--- a/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
@@ -33,18 +33,38 @@ if(_source_files)
   if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
     get_directory_property(_build_targets DIRECTORY ${CMAKE_SOURCE_DIR} BUILDSYSTEM_TARGETS)
     foreach(_target ${_build_targets})
-      get_property(_include_dirs
+      # Check if INCLUDE_DIRECTORIES is defined first in case the target is an INTERFACE
+      get_property(_include_dirs_defined
         TARGET ${_target}
         PROPERTY INCLUDE_DIRECTORIES
+        DEFINED
       )
+
+      if(_include_dirs_defined)
+        get_property(_include_dirs
+          TARGET ${_target}
+          PROPERTY INCLUDE_DIRECTORIES
+        )
+      else()
+        # Target might be an interface
+        get_property(_include_dirs
+          TARGET ${_target}
+          PROPERTY INTERFACE_INCLUDE_DIRECTORIES
+        )
+      endif()
 
       # Only append include directories that are from the package being tested
       # This accomplishes two things:
       #     1. Reduces execution time (less include directories to search)
       #     2. cppcheck will not check for errors in external packages
       foreach(_include_dir ${_include_dirs})
-        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_match ${_include_dir})
-        if(_is_match)
+        # TODO(jacobperron): Escape special regex characters in CMAKE_SOURCE_DIR
+        #                    Related CMake feature request: https://gitlab.kitware.com/cmake/cmake/issues/18409
+        # Check if include directory is a subdirectory of the source directory
+        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}/.*" _is_subdirectory ${_include_dir})
+        # Check if include directory is part of a generator expression (e.g. $<BUILD_INTERFACE:...>)
+        string(REGEX MATCH "^\\$<.*:${CMAKE_SOURCE_DIR}/.*>$" _is_genexp_subdirectory "${_include_dir}")
+        if(_is_subdirectory OR _is_genexp_subdirectory)
           list_append_unique(_all_include_dirs ${_include_dir})
         endif()
       endforeach()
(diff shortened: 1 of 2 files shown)


jacobperron:
CI (linter tests only):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6015/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2521/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4965/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5889/)

jacobperron:
FYI @Karsten1987

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -33,18 +33,36 @@ if(_source_files)
...
+
       # Only append include directories that are from the package being tested
       # This accomplishes two things:
       #     1. Reduces execution time (less include directories to search)
       #     2. cppcheck will not check for errors in external packages
       foreach(_include_dir ${_include_dirs})
-        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_match ${_include_dir})
-        if(_is_match)
+        # Check if include directory is a subdirectory of the source directory
+        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_subdirectory ${_include_dir})
```
dirk-thomas: This regex will also match sibling directories of e.g packages which start with the same name as this directory. So the regex should have a trailing shash before the wildcard.

Also this might need to consider platform specific path separators.

Same below.

jacobperron: This seems to be a way to do it 0472892
Open to suggestions.

nuclearsandwich: My impression was that CMake prefers paths to be stored with forward slashes internally and then converts them on platforms that need it but I can't find the relevant documentation.

jacobperron: Confirmed on a Windows machine that include dirs are cmake paths. I’ll remove the native path command.

jacobperron:
> when you run the next CI round for this can you include rosbag2? I think you'll need to instruct colcon to skip packages that require the ros1 bridge.

Done:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6022/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2528/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4974/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5898/)

jacobperron:
Regular CI (linter tests only):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6024/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2529/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4974/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5900/)

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -33,18 +33,36 @@ if(_source_files)
...
       # This accomplishes two things:
       #     1. Reduces execution time (less include directories to search)
       #     2. cppcheck will not check for errors in external packages
       foreach(_include_dir ${_include_dirs})
-        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}.*" _is_match ${_include_dir})
-        if(_is_match)
+        # Check if include directory is a subdirectory of the source directory
+        string(REGEX MATCH "^${CMAKE_SOURCE_DIR}/.*" _is_subdirectory ${_include_dir})
+        # Check if include directory is part of a generator expression (e.g. $<BUILD_INTERFACE:...>)
+        string(REGEX MATCH "^\\$<.*:${CMAKE_SOURCE_DIR}/.*>$" _is_genexp_subdirectory "${_include_dir}")
```
dirk-thomas: The `CMAKE_SOURCE_DIR` might contain characters which have special meaning in a regex. Unfortunately afaik there is no function in CMake to escape a string for that use case (see https://gitlab.kitware.com/cmake/cmake/issues/18409).

nuclearsandwich: I think the real world impact of this issue is narrowly scoped enough that the TODO in 600fa4c covers it.

A robust string library could probably get around the need for regexp here. Comparing the initia substring of the include directory to the CMAKE_SOURCE_DIR might be sufficient in the usual case but I don't have any idea yet what to do about the Generator Expressions.

jacobperron:
Looks like a regression, the errors have returned for cppcheck v1.86...

jacobperron:
For some reason `${_include_dirs_defined}` is always 0 (false), even though if the property is requested it works as expected:

https://github.com/ament/ament_lint/blob/63fc82d7249fd655c5313a83c82380a49f250b99/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake#L37-L41

jacobperron:
Through some experimentation, looks like we can use the `TYPE` property to check if a target is an `INTERFACE_LIBRARY`. Update to condition incoming...

jacobperron:
Windows (cppcheck v1.86) [Build Status](https://ci.ros2.org/job/test_ci_windows/42/)

jacobperron:
Test rosbag2_test_common:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/6029/)
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/5903/)

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -33,18 +33,26 @@ if(_source_files)
   if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
     get_directory_property(_build_targets DIRECTORY ${CMAKE_SOURCE_DIR} BUILDSYSTEM_TARGETS)
     foreach(_target ${_build_targets})
-      get_property(_include_dirs
-        TARGET ${_target}
-        PROPERTY INCLUDE_DIRECTORIES
-      )
+      # Include directories property is different for INTERFACE libraries
+      get_target_property(_target_type ${_target} TYPE)
```
nuclearsandwich: :tada: Glad you found a way to do this.

jacobperron:
For good measure:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6033/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2531/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4976/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5905/)

jacobperron:
Windows: [Build Status](https://ci.ros2.org/job/test_ci_windows/43/)

PR #122, opened 2019-01-22
use --extend-ignore for flake8 to keep default ignores

dirk-thomas:


dirk-thomas:
Ready for review.

ament_flake8/ament_flake8/configuration/ament_flake8.ini
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202,W504
+extend-ignore = C816,D100,D101,D102,D103,D104,D105,D106,D107,D203,D212,D404,I202
```
sloretz: It looks like this PR ignores `W503` (in `--ignore` defaults), where it wasn't ignored before. Is that intentional?

dirk-thomas: Unfortunately I don't see a way to additionally enable `W503` without without either explicitly listing each and every possible code in `--select` or the inverse in `-ignore`.

Using `--extend-ignore` has the benefit that future code disabled by default won't affect us anymore. The downside is we don't check for `W503` anymore.

PR #125, opened 2019-02-12
use globally defined varible for cppcheck include dirs

Karsten1987:
introduces a new variable `AMENT_CMAKE_CPPCHECK_INCLUDE_DIRS` which can be set externally to configure cppcheck's include dirs.

the new variable can then be used externally in projects as such:

```
if(BUILD_TESTING)
  find_package(ament_cmake_gmock REQUIRED)
  find_package(ament_lint_auto REQUIRED)
  find_package(test_msgs REQUIRED)
  find_package(rosbag2_test_common REQUIRED)

  set(AMENT_CMAKE_CPPCHECK_INCLUDE_DIRS ${rclcpp_INCLUDE_DIRS})

  [...]
```

connects to the CI warnings found here: https://github.com/ros2/rosbag2/pull/86

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
@@ -29,6 +29,10 @@ if(_source_files)
 
   # Get include paths for added targets
   set(_all_include_dirs "")
+  if(DEFINED ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS)
+    list(APPEND _all_include_dirs ${ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS})
+  endif()
+
   # BUILDSYSTEM_TARGETS only supported in CMake >= 3.7
   if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
     get_directory_property(_build_targets DIRECTORY ${CMAKE_SOURCE_DIR} BUILDSYSTEM_TARGETS)
@@ -59,5 +63,6 @@ if(_source_files)
     endforeach()
   endif()
 
+  message(STATUS "Configured cppcheck include dirs: ${_all_include_dirs}")
   ament_cppcheck(INCLUDE_DIRS ${_all_include_dirs})
 endif()
--- a/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
@@ -15,6 +15,10 @@
 #
 # Add a test to perform static code analysis with cppcheck.
 #
+# The include dirs for cppcheck to consider can either be set by the function
+# parameter 'INCLUDE_DIRS` or by a global variable called
+# 'ament_cmake_cppcheck_ADDITIONAL_INCLUDE_DIRS'.
+#
 # :param TESTNAME: the name of the test, default: "cppcheck"
 # :type TESTNAME: string
 # :param LANGUAGE: the language argument for cppcheck, either 'c' or 'c++'


ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -29,6 +29,10 @@ if(_source_files)
 
   # Get include paths for added targets
   set(_all_include_dirs "")
+  if(DEFINED AMENT_CMAKE_CPPCHECK_INCLUDE_DIRS)
```
dirk-thomas: I would suggest to use lowercase for the `AMENT_CMAKE_CPPCHECK` to match the package name and other variables in style.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
```diff
@@ -29,6 +29,10 @@ if(_source_files)
...
+    list_append_unique(_all_include_dirs ${AMENT_CMAKE_CPPCHECK_INCLUDE_DIRS})
```
dirk-thomas: `list(APPEND ...)` should be sufficient here since the list is empty and the variable provided by the user which has control about avoiding duplicates in the first place.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
Karsten1987: now that I think about, maybe a completely different name could be good. Might be confusing with the default c++ like include dirs.
what about `ament_cmake_cppcheck_EXTERNAL_DIRS`?

jacobperron: `ament_cmake_cppcheck_EXTERNAL_INCLUDE_DIRS`?

dirk-thomas:
For your example usage of this new API what are you intending to set the variable to? Are these paths actually external?

Karsten1987:
yes, they are meant to be external. AFAIK, the cmake logic implemented in this file gets all local/internal include dirs and specifically excludes external ones. @jacobperron is my understanding here correct?

jacobperron:
> AFAIK, the cmake logic implemented in this file gets all local/internal include dirs and specifically excludes external ones. @jacobperron is my understanding here correct?

Correct.

---

The affected package is `rosbag2_transport` (ros2/rosbag2#86). It is [using a macro from rclcpp](https://github.com/ros2/rosbag2/blob/733821089d93b30fa06455d7d6dd1a6291c83e74/rosbag2_transport/src/rosbag2_transport/generic_subscription.hpp#L36), which `cppcheck` complains about as an "unknown" macro. By providing `cppcheck` a reference to the `rclcpp` header containing the macro, it can resolve the includes and the error goes away.

I expect this issue is likely to affect other packages that similarly try to use the same macros from `rclcpp` or potentially macros from other libraries. I'm not sure if there's a way to handle this as part `ament_auto_lint`, but in the meantime any affected packages can make use of the variable introduced in this PR.

ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
dirk-thomas: 650de6975bb269a2716343bbf39901c92b889dd9

dirk-thomas:
I am not convinced that "external" is a good prefix here. There is no reason that the path needs to be external. It can literally point anywhere - e.g. into the packages build directory. Therefore I would suggest a different prefix, e.g.: `ADDITIONAL` or `EXTRA`.

dirk-thomas:
@Karsten1987 also please use "Sign-Off-By" for your commits to pass the DCO check.

Karsten1987:
changed to `ADDITIONAL_INCLUDE_DIRS`. Commit 104cc2c is signed.

dirk-thomas:
> changed to `ADDITIONAL_INCLUDE_DIRS`. Commit [05c794c](https://github.com/ament/ament_lint/commit/05c794c8dcc5ecae0bc425e6c842c9a638ce0857) is signed.

:+1: 

Please update the [docblock](https://github.com/ament/ament_lint/blob/4f9bfc00106496eeafa592bc5dfbc33b052d643d/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake#L15-L28) to mention the new variable.

Karsten1987:
updated docblock mentioning how to set the include dirs for cppcheck

Karsten1987:
rebased for DCO compliance.

Issue #132, opened 2019-03-20
Allow excluding checkers from ament_lint_auto_find_test_dependencies


jpsamper2009:
## Motivation
- As mentioned in https://github.com/ros2/rosbag2/pull/86#issuecomment-461099130, it is possible to exclude specific checkers by setting `<package>_FOUND` to `TRUE`: e.g.

```cmake
find_package(ament_lint_auto REQUIRED)
# Disable auto copyright
set(ament_cmake_copyright_FOUND TRUE)
# Disable auto cpplint
set(ament_cmake_cpplint_FOUND TRUE)
ament_lint_auto_find_test_dependencies()
```

- However, if you still want to run the checkers (with a different configuration or on a subset of files, for example), the cmake code above becomes:

```cmake
find_package(ament_lint_auto REQUIRED)
# Disable auto copyright
set(ament_cmake_copyright_FOUND TRUE)
# Disable auto cpplint
set(ament_cmake_cpplint_FOUND TRUE)
ament_lint_auto_find_test_dependencies()

# Reenable copyright
set(ament_cmake_copyright_FOUND FALSE)
find_package(ament_cmake_copyright)
ament_copyright(${ONLY_SOME_FILES})

# Reenable cpplint
set(ament_cmake_cpplint_FOUND FALSE)
find_package(ament_cmake_cpplint_FOUND)
ament_cpplint(FILTERS "-readability/inheritance")
```

- Adding and `EXCLUDE` option in `ament_lint_auto_find_test_dependencies` could make the code above cleaner: e.g.

```cmake
find_package(ament_lint_auto REQUIRED)
ament_lint_auto_find_test_dependencies(
  EXCLUDE
  ament_cmake_copyright
  ament_cmake_cpplint
)

find_package(ament_cmake_copyright)
ament_copyright(${ONLY_SOME_FILES})

find_package(ament_cmake_cpplint)
ament_cpplint(FILTERS "-readability/inheritance")
```

- My thought is that we can basically move the `set()` commands into the [macro](https://github.com/ament/ament_lint/blob/master/ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake)

## Goals

- Simplify the `CMakeLists.txt` file of packages which have a checkers or files diverging from the default checkers
- Make sure that new files are automatically checked, such that they need to be exclude explicitly
- Make sure that new linters (e.g. `xmllint`) automatically get added

## Relate issues

[...]

PR #133, opened 2019-03-21
Allow excluding checkers from ament_lint_auto

jpsamper2009:
This PR implements the enhancement proposed in #132. It consists of two commits:

1. A bug fix to the way `ament_lint_auto_find_test_dependencies` detected unused arguments
2. Adding AMENT_LINT_AUTO_EXCLUDE variable to `ament_execute_extension`

- See https://github.com/ament/ament_cmake/pull/165 also
- Example: ros/class_loader#123

## How to test
- Take a package like `rclcpp` which uses `ament_lint_auto_find_test_dependencies` (https://github.com/ros2/rclcpp/blob/master/rclcpp/CMakeLists.txt#L140) :

1. Introduce a couple of style error in a source file, run `colcon build`, `colcon test --packages-select rclcpp` and see that the `cpplint` and `uncrustify` tests fail

2. Exclude one of the checkers:

```cmake
list(APPEND AMENT_LINT_AUTO_EXCLUDE
  ament_cmake_uncrustify
)
```

- [ ] Run `colcon build --packages-select rclcpp`, `colcon test --packages-select rclcpp` and see that the uncrustify test does not exist any more

3. Find and add the `ament_cmake_uncrustify` check explicitly, excluding the file with the style errors:

```cmake
list(APPEND AMENT_LINT_AUTO_EXCLUDE
  ament_cmake_uncrustify
)
find_package(ament_cmake_uncrustify)

set(good_sources "${rclcpp_SRCS}")
list(REMOVE_ITEM good_sources <bad-file>)
ament_uncrustify(${good_sources})
```
- [ ] `uncrustify` should succeed

4. Exclude `ament_cmake_cpplint`
```cmake
list(APPEND AMENT_LINT_AUTO_EXCLUDE
  ament_cmake_uncrustify
  ament_cmake_cpplint
)
find_package(ament_cmake_uncrustify)

set(good_sources "${rclcpp_SRCS}")
list(REMOVE_ITEM good_sources <bad-file>)
ament_uncrustify(${good_sources})
```

- [ ] `cpplint` test does not exist anymore

5. Add the `cpplint` test (before `ament_lint_auto`), but use the `FILTERS` option to suppress the error message:

```cmake
find_package(ament_cmake_cpplint)
ament_cpplint(FILTERS "-foo")

list(APPEND AMENT_LINT_AUTO_EXCLUDE
  ament_cmake_uncrustify
  ament_cmake_cpplint
)

find_package(ament_cmake_uncrustify)
set(good_sources "${rclcpp_SRCS}")
list(REMOVE_ITEM good_sources <bad-file>)
ament_uncrustify(${good_sources})
```

- [ ] `cpplint` test succeeds

### DIFF
--- a/ament_lint_auto/cmake/ament_lint_auto_package_hook.cmake
+++ b/ament_lint_auto/cmake/ament_lint_auto_package_hook.cmake
@@ -12,4 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ament_execute_extensions(ament_lint_auto)
+# Use the AMENT_LINT_AUTO_EXCLUDE variable to exclude linter packages from
+# running using the ament_lint_auto hook.
+# It is still possible, however, to use the linter by calling it explicitly.
+#
+# :variable AMENT_LINT_AUTO_EXCLUDE: List of strings (package names)
+
+ament_execute_extensions(ament_lint_auto EXCLUDE "${AMENT_LINT_AUTO_EXCLUDE}")


ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -21,6 +21,7 @@
 # @public
 #
 macro(ament_lint_auto_find_test_dependencies)
+  cmake_parse_arguments(ARG "" "" "EXCLUDE" ${ARGN})
```
dirk-thomas: Since this is a macro please prefix all "local" variables with an underscore: `_`.

The new option also needs to be documented in the docblock.

Doesn't the following check have to be remove to make it even continue when `EXCLUDE` is passed? Have you tried to use this patch with the new `EXCLUDE` option? I would suggest to also update at least one of the existing cases where we skip a specific linter by setting the `FOUND` flag to `FALSE` before to use this new option to demonstrate this works.

ament_cmake_clang_format/ament_cmake_clang_format-extras.cmake
```diff
@@ -18,5 +18,7 @@ find_package(ament_cmake_test QUIET REQUIRED)
 
 include("${ament_cmake_clang_format_DIR}/ament_clang_format.cmake")
 
-ament_register_extension("ament_lint_auto" "ament_cmake_clang_format"
-"ament_cmake_clang_format_lint_hook.cmake")
+if(REGISTER_AMENT_LINT_AUTO_EXTENSIONS)
+  ament_register_extension("ament_lint_auto" "ament_cmake_clang_format"
+    "ament_cmake_clang_format_lint_hook.cmake")
+endif()
```
dirk-thomas: Please keep the newline at the end of the file.

Same below.

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -30,11 +31,28 @@ macro(ament_lint_auto_find_test_dependencies)
...
+  set(EXCLUDED ${ARG_EXCLUDE})
+  foreach(_exclusion ${ARG_EXCLUDE})
+    # If a package has already been found, we don't want to
+    # set <package>_FOUND to FALSE at the end
+    if(${_exclusion}_FOUND)
+      list(REMOVE_ITEM EXCLUDED ${_exclusion})
+    else()
+      set(${_exclusion}_FOUND TRUE)
+    endif()
+  endforeach()
```
dirk-thomas: Instead of faking the `FOUND` flag here why not skip `_dep`s in the following `foreach` if they are listed in `ARG_EXCLUDE`?

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
jpsamper2009: @dirk-thomas Did you mean "...where we skip a specific linter by setting the FOUND flag to **TRUE**"? e.g.  https://github.com/ament/ament_index/blob/master/ament_index_cpp/CMakeLists.txt#L43

---

The check doesn't fail, but you are right that I should check for ARG_UNPARSED_ARGUMENTS. It turns out that macro arguments are not variables which means they don't work with `if()` (https://cmake.org/cmake/help/v3.0/command/if.html):

> `if(<variable>)`
    True if the variable is defined to a value that is not a false constant. False otherwise. (Note macro arguments are not variables.)

The following would've caught the error we are expecting:

```cmake
set(FOO ${ARGN})
if(FOO)
  message(FATAL_ERROR ...)
endif()
```

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
jpsamper2009: @dirk-thomas Skipping `find_package` only works if the package is a direct dependency (and is therefore in `*_TEST_DEPENDS`); however, if you are using a meta package like `ament_lint_common`, `find_package(ament_cmake_cppcheck)` will not be skipped

ament_cmake_clang_format/ament_cmake_clang_format-extras.cmake
jpsamper2009: I didn't see that. Good catch.

jpsamper2009:
@dirk-thomas I think I've addressed all your comments. I have created the following PR (https://github.com/ament/ament_index/pull/31) as an example of using the argument

dirk-thomas:
I don't think the current approach with the global flag works well in a kind of scenarios. E.g. a package finds on of the linters before (when the new global flag isn't set yet).

How about just storing the excluded linter names in a global variable and instead of blindly triggering the extension point with `ament_execute_extensions(ament_lint_auto)` iterate over the registered extensions and skip the ones from the exclude list? That avoids all custom magic and find package handling.

jpsamper2009:
@dirk-thomas Are you suggesting that we replace the call to [`ament_execute_extensions(ament_lint_auto)`](https://github.com/ament/ament_lint/blob/master/ament_lint_auto/cmake/ament_lint_auto_package_hook.cmake) with 

the contents of [`ament_execute_extensions.cmake`](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/cmake/core/ament_execute_extensions.cmake) + some logic for excluded packages

OR

to modify `ament_execute_extensions.cmake` to look for a global variable with a list of excluded packages?

dirk-thomas:
Copying the logic is not great. Also making this function rely on global state doesn't seem nice. How about adding an optional keyword argument to `ament_execute_extensions` to skip a set of given package names?

jpsamper2009:
@dirk-thomas The issue with relying on `ament_execute_extensions` is that the tests would still be registered but not executed, which means that you would still not be able to register a test a customer test because there would already be a test with that name

dirk-thomas:
If a certain linter is being excluded it shouldn't define any test so I don't see why the test should be registered in that case? Also each CMake function adding a linter test has a `TESTNAME` keyword argument to define your own test name (though based on the previous sentence I don't think this should be necessary if the linter is excluded).

jpsamper2009:
@dirk-thomas I've pushed a new proposed solution based on our discussions above. This requires changes to `ament_cmake` too: https://github.com/ament/ament_cmake/pull/165

I have tested it with our own packages, but I don't seem to find a good package in ros2 to show this working. What would be the best way to provide an example?

dirk-thomas:
`grep -R "_FOUND TRUE)"` through the ROS 2 workspace finds a few candidates - either of them would be a good example:

* [class_loader](https://github.com/ros/class_loader/blob/687424f6112067cd236c0c8f5e7de2300403588a/CMakeLists.txt#L56-L57)
* [kdl_parser](https://github.com/ros2/kdl_parser/blob/480c320bbd86264ab5ce0547a8962dc2b5953b8f/kdl_parser/CMakeLists.txt#L56-L58)
* [test_security](https://github.com/ros2/system_tests/blob/a5d2c6daffe1ffd9baf39186f3b442a38fcffb35/test_security/CMakeLists.txt#L47-L49)

jpsamper2009:
@dirk-thomas Sounds good. See https://github.com/ros/class_loader/pull/123

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -17,15 +17,24 @@
 #
 # All found package names are appended to the
 # ``${PROJECT_NAME}_FOUND_TEST_DEPENDS`` variables.
+# :param EXCLUDE: List of linter packages that should be skipped
+# :type EXCLUDE: list of strings
+#
+# :outvar <project_name>_FOUND_TEST_DEPENDS: All found package names
```
dirk-thomas: This seems to be redundant with line 18/19.

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -17,15 +17,24 @@
...
-  if(ARGN)
+  cmake_parse_arguments(_ARG "" "" "EXCLUDE" ${ARGN})
+  if(_ARG_UNPARSED_ARGUMENTS)
     message(FATAL_ERROR "ament_lint_auto_find_test_dependencies() called with "
-      "unused arguments: ${ARGN}")
+      "unused arguments: ${_ARG_UNPARSED_ARGUMENTS}")
   endif()
 
+  # Set global variable used by ament_execute_extensions
+  set(AMENT_LINT_AUTO_EXCLUDE ${_ARG_EXCLUDE})
```
dirk-thomas: Maybe only `list(APPEND` here to allow the caller to populate the list outside.

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -17,15 +17,24 @@
 #
 # All found package names are appended to the
 # ``${PROJECT_NAME}_FOUND_TEST_DEPENDS`` variables.
+# :param EXCLUDE: List of linter packages that should be skipped
+# :type EXCLUDE: list of strings
```
dirk-thomas: Based on the docs I would expect that the listed packages are also not being `find_package()`-ed.

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -17,15 +17,24 @@
...
+#
+# :outvar <project_name>_FOUND_TEST_DEPENDS: All found package names
+# :outvar AMENT_LINT_AUTO_EXCLUDE: List of excluded packages
```
dirk-thomas: The doc should mention that all excluded packages are added to this variable and that the variable is being passed to `ament_execute_extensions`.

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
jpsamper2009: We would have to go back to the messy `set(<package>_FOUND TRUE)` solution that I first proposed (https://github.com/ament/ament_lint/pull/133#discussion_r268344474)... maybe it's better to just let users set the variable outside of `ament_lint_auto_find_test_dependencies`?

I'm just not sure where to document that this variable exists and how to use it.

dirk-thomas: I keep forgetting that the test dependencies are not necessarily immediately the linter packages. :thinking:

Since the function `ament_lint_auto_find_test_dependencies` doesn't do anything beside setting a global variable I would lean towards not adding the argument to it.

> maybe it's better to just let users set the variable outside of `ament_lint_auto_find_test_dependencies`?
 > I'm just not sure where to document that this variable exists and how to use it.

Let's do that. Since there is no direct function being called there is no good docblock to put the docs. Maybe just put it into `ament_lint_auto_package_hook.cmake` where it is going to be used.

jpsamper2009:
@dirk-thomas This is ready for a review

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
```diff
@@ -21,9 +21,10 @@
 # @public
 #
 macro(ament_lint_auto_find_test_dependencies)
-  if(ARGN)
+  set(_ARGN "${ARGN}")
+  if(_ARGN)
     message(FATAL_ERROR "ament_lint_auto_find_test_dependencies() called with "
-      "unused arguments: ${ARGN}")
+      "unused arguments: ${_ARGN}")
```
dirk-thomas: These changes look unrelated. Why did you change them?

ament_lint_auto/cmake/ament_lint_auto_package_hook.cmake
```diff
@@ -12,4 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ament_execute_extensions(ament_lint_auto)
+# Use the AMENT_LINT_AUTO_EXCLUDE variable to exclude linter packages from
+# running using the ament_lint_auto hook. It is still possible, however, to
```
dirk-thomas: Each sentence should start on a new line (see https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#markdown-restructured-text-docblocks).

ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
jpsamper2009: That's related to this comment: https://github.com/ament/ament_lint/pull/133#discussion_r268343524

I'll submit them separately - this PR originally had a second change to these lines where it checked ARG_UNPARSED_ARGUMENTS

jpsamper2009: https://github.com/ament/ament_lint/pull/136

jpsamper2009:
@dirk-thomas One last review, hopefully :smile:

dirk-thomas:
> One last review, hopefully :smile:

All good :+1: Thanks for iterating on it.

dirk-thomas:
I added a version dependency for this since the new argument is only available as of version 0.7.0: 8b6c495c37ae2964565b3d2ab2145882445c0dbc

PR #135, opened 2019-03-28
use yaml.safe_load instead of deprecated yaml.load

mikaelarguedas:
Similar to https://github.com/colcon/colcon-metadata/pull/14

Note: there may be a similar fix to apply on the `yaml.dump` side. But I didn't face deprecation warnings so far so focused this change on the load side.


dirk-thomas:
Thanks for the patch.

PR #136, opened 2019-04-11
Fix unused-arg check in auto find test deps

jpsamper2009:
Arguments to a macro are not variables, so it's not
possible to do 'if(ARGN)' to check for arguments;
however, copying ARGN to a variable works.

See https://cmake.org/cmake/help/v3.5/command/if.html (emphasis mine):

> if(<variable|string>)
True if given a variable that is defined to a value that is not a false constant. False otherwise. **(Note macro arguments are not variables.)**

- See also ament/ament_cmake#167

### DIFF
--- a/ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
+++ b/ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake
@@ -21,9 +21,10 @@
 # @public
 #
 macro(ament_lint_auto_find_test_dependencies)
-  if(ARGN)
+  set(_ARGN "${ARGN}")
+  if(_ARGN)
     message(FATAL_ERROR "ament_lint_auto_find_test_dependencies() called with "
-      "unused arguments: ${ARGN}")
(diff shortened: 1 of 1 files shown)


dirk-thomas:
That contradicts what I see with this tiny snippet:

```
function(func)
  if(ARGN)
    message("if ${ARGN}")
  else()
    message("else")
  endif()
endfunction()

func(arg)
func()
```

Can you please double check with that example and comment on your CMake version if it fails for you.

jpsamper2009:
@dirk-thomas In your example, you're using a function; this behavior applies to macros.

I'm using CMake 3.5.1

dirk-thomas:
And another nasty surprise brought to you by CMake :disappointed: Then the patch is good as is.

Since the same pattern is present in a few other macros (`ament_cmake_gmock`, `ament_cmake_gtest`, `ament_cmake_nose`) could you maybe fix those cases too?

jpsamper2009:
@dirk-thomas Done. Link to the other PR is in the description.

dirk-thomas:
Thanks for the fix.

PR #140, opened 2019-05-07
Fix deprecation warning using raw str

sloretz:
This fixes a deprecation warning when building `ament_lint_cmake` on windows 10 using python 3.7.3

```
C:\ros2_ws\build\ament_lint_cmake\build\bdist.win-amd64\egg\ament_lint_cmake\cmakelint.py:124: DeprecationWarning: invalid escape sequence \.
  package = re.sub('^Find(.*)\.cmake', lambda m: m.group(1), package)
C:\ros2_ws\build\ament_lint_cmake\build\bdist.win-amd64\egg\ament_lint_cmake\cmakelint.py:335: DeprecationWarning: invalid escape sequence \.
  name_match = re.match('Find(.*)\.cmake', os.path.basename(filename))
```

### DIFF
--- a/ament_lint_cmake/ament_lint_cmake/cmakelint.py
+++ b/ament_lint_cmake/ament_lint_cmake/cmakelint.py
@@ -121,7 +121,7 @@ class _CMakePackageState(object):
 
     def _GetExpected(self, filename):
         package = os.path.basename(filename)
-        package = re.sub('^Find(.*)\.cmake', lambda m: m.group(1), package)
+        package = re.sub(r'^Find(.*)\.cmake', lambda m: m.group(1), package)
         return package.upper()
 
     def Done(self, filename, errors):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Have these changes been contributed upstream? If not I would recommend that and once they have been accepted we can pull an updated version.

sloretz:
I didn't realize this file was taken from another repo. It looks like `cmakelint` is [available on pip](https://pypi.org/project/cmakelint/), any reason not to use that?

dirk-thomas:
> It looks like `cmakelint` is available on PyPI, any reason not to use that?

We need it to be a Debian package in order to depend on it in a Debian package.

hidmic:
Hmm, isn't this already in [`master`](https://github.com/ament/ament_lint/commit/45d806998a2c72b85b906b70f3dca3ad71e7fc61)?

sloretz:
Replaced by #144

PR #142, opened 2019-05-09
handle BOM properly

dirk-thomas:
Needed for ros2/rcl_interfaces#79


hidmic:
It works just fine locally, running CI now for both PRs now:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6852/)

PR #147, opened 2019-07-03
Bring clang-format in line with developer guide

rotu:
Previously, ament_clang_format's defaults were very divergent from ament_uncrustify. Now the default config file has been rewritten to match the developer guide.
https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#id26

### DIFF
--- a/ament_clang_format/ament_clang_format/configuration/.clang-format
+++ b/ament_clang_format/ament_clang_format/configuration/.clang-format
@@ -1,87 +1,12 @@
+---
 Language: Cpp
-# BasedOnStyle: Google
-AccessModifierOffset: -1
-AlignAfterOpenBracket: Align
-AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: false
-AlignEscapedNewlinesLeft: true
-AlignOperands: true
-AlignTrailingComments: true
-AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortBlocksOnASingleLine: false
-AllowShortCaseLabelsOnASingleLine: false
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please keep the list of option in alphabetical order. That will hopefully also improve the readability of the diff.

Regarding the deleted entries: did  you delete them because they are the same as in the Google style? Not sure if it would make sense to keep them explicitly for easier review. If not, I think having two separate commits - one for changing values and one for removing values which are the same as the default - would be good.

rotu:
Thanks for the feedback. I had them in the order they were mentioned in the dev guide, but alphabetical makes more sense.

Yes, I deleted extra entries because they are the same as in the Google style, but also because the ROS2 dev guide didn't mention them, and I wanted this to be as faithful as possible to that document.

dirk-thomas:
Have you tried to use clang format with the updated configuration on the ROS 2 code base? Currently clang format is excluded by default (see https://github.com/ament/ament_lint/blob/40cf85e6804c30247163f8c8a6f19f7afd177226/ament_lint_common/package.xml#L15). In order for CI to use it it needs to be enabled.

rotu:
Yes, I use clang-format with that config on my code. No, I can't fully use it on the ROS2 code base. It still cannot be enabled by default since it does not agree with `ament_uncrustify` in all cases.

rotu:
@dirk-thomas What's the status of this PR? Are there changes you're waiting on?

dirk-thomas:
Since this isn't used anywhere I am happy to merge it as is.

PR #153, opened 2019-07-17
Add custom config file support for clang-format

JShep1:
Currently only the `.clang-format` file in the `configuration/` directory can be used.  This update adds support for custom `.clang-format` files placed in the directory from which `ament_clang_format` is called.

Signed-off-by: John Shepherd <johnshepherd96@yahoo.com>

### DIFF
--- a/ament_cmake_clang_format/cmake/ament_clang_format.cmake
+++ b/ament_cmake_clang_format/cmake/ament_clang_format.cmake
@@ -25,7 +25,7 @@
 # 
 # :param TESTNAME: the name of the test, default: "clang_format"
 # :type TESTNAME: string
-# :param CONFIG_FILE: the name of the configuration file for clang-format to consider
+# :param CONFIG_FILE: the path of the configuration file for clang-format to consider
 # :type CONFIG_FILE: string
 # :param ARGN: the files or directories to check
 # :type ARGN: list of strings
--- a/ament_cmake_clang_format/cmake/ament_clang_format.cmake
+++ b/ament_cmake_clang_format/cmake/ament_clang_format.cmake
@@ -15,9 +15,13 @@
 #
 # Add a test to check the code for compliance with clang_format.
 #
-# The configuration file for clang-format to use can be either set by
-# the function parameter 'CONFIG_PATH' or by a global variabled called
+# The default configuration file used for clang-format is located at 
+# configuration/.clang-format within the ament_clang_format directory
+# The default configuration file can be either overridden by the
+# argument 'CONFIG_FILE' or by a global variable named
 # 'ament_cmake_clang_format_CONFIG_FILE'
+# The 'CONFIG_FILE' argument takes priority over 
+# 'ament_cmake_clang_format_CONFIG_FILE' if both are defined
 # 
 # :param TESTNAME: the name of the test, default: "clang_format"
 # :type TESTNAME: string
@@ -44,6 +48,8 @@ function(ament_clang_format)
   list(APPEND cmd ${ARG_UNPARSED_ARGUMENTS})
   if(ARG_CONFIG_FILE)
     set(APPEND cmd "--config" "${ARG_CONFIG_FILE}")
+  elseif(DEFINED ament_cmake_clang_format_CONFIG_FILE)
+    set(APPEND cmd "--config" "${ament_cmake_clang_format_CONFIG_FILE}")
   endif()
   
   file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/ament_clang_format")
--- a/ament_cmake_clang_format/cmake/ament_clang_format.cmake
+++ b/ament_cmake_clang_format/cmake/ament_clang_format.cmake
@@ -15,15 +15,21 @@
 #
 # Add a test to check the code for compliance with clang_format.
 #
+# The configuration file for clang-format to use can be either set by
+# the function parameter 'CONFIG_PATH' or by a global variabled called
+# 'ament_cmake_clang_format_CONFIG_FILE'
+# 
 # :param TESTNAME: the name of the test, default: "clang_format"
 # :type TESTNAME: string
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Using a file in the cwd is kind of opaque. You have to know that this feature exists in order to use it.

I would suggest to instead add a command line option to the CLI as well as a keyword argument and global variable to the CMake API.

JShep1:
Added suggested CMake changes

ament_cmake_clang_format/cmake/ament_clang_format.cmake
```diff
@@ -15,15 +15,21 @@
 #
 # Add a test to check the code for compliance with clang_format.
 #
+# The configuration file for clang-format to use can be either set by
```
dirk-thomas: > ... can be either set by ...

Proposing: ... can be either overridden by ...

Maybe add a sentence before that mentions the configuration file used by default?

ament_cmake_clang_format/cmake/ament_clang_format.cmake
```diff
@@ -15,15 +15,21 @@
...
+# the function parameter 'CONFIG_PATH' or by a global variabled called
```
dirk-thomas: > the function parameter ...

Proposing: the argument ...

> 'CONFIG_PATH'

Should be: 'CONFIG_FILE'

> variabled

Spelling: variable

> called

Proposing: named

ament_cmake_clang_format/cmake/ament_cmake_clang_format_lint_hook.cmake
```diff
@@ -23,6 +23,10 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
   "*.hxx"
 )
 if(_source_files)
+  if(DEFINED ament_cmake_clang_format_CONFIG_FILE)
+    set(config_file_path ${ament_cmake_clang_format_CONFIG_FILE})
+  endif()
```
dirk-thomas: In the current form the global variable is only considered by this auto-lint hook. The function doc suggests that it also affect any manual call of `ament_clang_format`.

Depending on which behavior is desired either the docs have to be changed or code has to be modified to move the usage of the global variable into the `ament_clang_format` function. I am leaning slightly toward the second one - since it is easier to document.

`cppcheck` might need the same fix since it looks like the pattern there is the same.

JShep1: If there is a case in which both the `CONFIG_FILE` and `ament_cmake_clang_format_CONFIG_FILE` are defined, which of the two should be used?  I'm leaning towards `CONFIG_FILE` given that's the most explicit config path specified by the user at the time of calling

ament_cmake_clang_format/cmake/ament_clang_format.cmake
JShep1: Thanks for pointing these out

ament_cmake_clang_format/cmake/ament_clang_format.cmake
JShep1: Just added the suggested change along with a sentence specifying the default configuration file

ament_cmake_clang_format/cmake/ament_cmake_clang_format_lint_hook.cmake
dirk-thomas: > If there is a case in which both `CONFIG_FILE` and `ament_cmake_clang_format_CONFIG_FILE` are defined

Could be.

> I'm leaning towards CONFIG_FILE given that's the most explicit config path specified by the user at the time of calling

:+1:

ament_cmake_clang_format/cmake/ament_cmake_clang_format_lint_hook.cmake
```diff
@@ -23,6 +23,10 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
...
+
   message(STATUS "Added test 'clang_format' to check C / C++ code style")
-  ament_clang_format()
+  ament_clang_format(CONFIG_FILE ${config_file_path})
```
dirk-thomas: This doesn't need to pass a `CONFIG_FILE` anymore.

JShep1: I was thinking that but was unsure if I should make the change or not - thanks

ament_cmake_clang_format/cmake/ament_clang_format.cmake
```diff
@@ -15,15 +15,25 @@
...
+# configuration/.clang-format within the ament_clang_format directory
+# The default configuration file can be either overridden by the
+# argument 'CONFIG_FILE' or by a global variable named
+# 'ament_cmake_clang_format_CONFIG_FILE'
+# The 'CONFIG_FILE' argument takes priority over 
+# 'ament_cmake_clang_format_CONFIG_FILE' if both are defined
+# 
 # :param TESTNAME: the name of the test, default: "clang_format"
 # :type TESTNAME: string
+# :param CONFIG_FILE: the name of the configuration file for clang-format to consider
```
dirk-thomas: Last nitpick: the *path* of the configuration file

wjwwood:
Looks like this introduced some linter failures on the nightly CI:

https://ci.ros2.org/view/nightly/job/nightly_linux_release/1247/testReport/(root)/projectroot/lint_cmake/

@JShep1 can you please fix those up? Also, it's preferable if you run CI even for small changes like this before merging. Thanks!

wjwwood:
Btw, If you need help running the tests locally, or configuring CI so it's as quick as possible while still covering the change, let me know.

JShep1:
Resolved in https://github.com/ament/ament_lint/pull/162

PR #155, opened 2019-07-19
Add clang tidy to ament linters

JShep1:
Adds `ament_clang_tidy` functionality to ament with default Google configuration checks enabled.


ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env python3
+
+# Copyright 2014-2016 Open Source Robotics Foundation, Inc.
```
clalancette: Nit: 2019.  Ditto everywhere in this PR.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -0,0 +1,290 @@
...
+        start_time = time.time()
+
+    files = get_files(args.paths, extensions)
+    if not files:
+        print('No files found', file=sys.stderr)
+        return 1
+
+    bin_names = [
+        'clang-tidy',
+        'clang-tidy-6.0',
```
clalancette: You may want to add `clang-tidy-3.8` here, as that is the version on Xenial.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -0,0 +1,290 @@
...
+        xml = get_xunit_content(report, testname, time.time() - start_time)
+        path = os.path.dirname(os.path.abspath(args.xunit_file))
+        if not os.path.exists(path):
+            os.makedirs(path)
+        with open(args.xunit_file, 'w') as f:
+            f.write(xml)
+    return
+
+
+def find_executable(file_names):
```
clalancette: Hm.  The duplication of `find_executables`, `get_files`, `get_xunit_content`, etc. with other ament linters is pretty unfortunate.  However, I see that is how we have done it for all of the linters, and I think it deserves a different patch.  I'll request that you open an issue to investigate how to share more code between them.

ament_clang_tidy/doc/index.rst
```diff
@@ -0,0 +1,28 @@
+ament_clang_tidy
+==================
+
+Checks the code style of C / C++ source files using `ClangTidy
```
clalancette: Nit: it is `Clang-Tidy` in their documentation.

ament_clang_tidy/doc/index.rst
```diff
@@ -0,0 +1,28 @@
...
+<http://clang.llvm.org/extra/clang-tidy/>`_.
+Files with the following extensions are being considered:
```
clalancette:
```suggestion
Files with the following extensions are considered:
```

ament_clang_tidy/package.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ament_clang_tidy</name>
+  <version>6.0.0</version>
```
clalancette: I'd reduce this version number to 1.0.0, since this is new.  Ditto for setup.py.

ament_clang_tidy/package.xml
```diff
@@ -0,0 +1,22 @@
...
+  <description>
+    The ability to check code against style conventions using
+    clang-tidy and generate xUnit test result files.</description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
```
clalancette: Make yourself the maintainer; I think Dirk already has enough :).  Ditto for setup.py

ament_clang_tidy/ament_clang_tidy/main.py
dirk-thomas: Why is the version from Xenial relevant here? Our current target platform is Bionic which ships 6.0.

ament_clang_tidy/package.xml
dirk-thomas: The version number must match the version of the other packages in this repo.

ament_clang_tidy/package.xml
dirk-thomas: +:100:

ament_clang_tidy/ament_clang_tidy/main.py
clalancette: For ROS 2, yes.  For another project that we are considering using this on, they still target Xenial.  The other linters also generally support older versions, so it seems like we should follow that trend.

ament_clang_tidy/package.xml
clalancette: Oops, yeah, forgot about this.  Never mind this comment then :).

ament_clang_tidy/ament_clang_tidy/main.py
dirk-thomas: I would expect significant differences between version 6.0 and 3.8. So I am not sure if supporting such wide spread is feasible.

ament_clang_tidy/ament_clang_tidy/main.py
JShep1: I was thinking the same thing implementing this section - I'll open the issue shortly

ament_clang_tidy/setup.py
```diff
@@ -0,0 +1,37 @@
...
+    name='ament_clang_tidy',
+    version='6.0.0',
+    packages=find_packages(exclude=['test']),
+    install_requires=['setuptools', 'pyyaml'],
+    package_data={'': [
+        'configuration/.clang-tidy',
+    ]},
+    zip_safe=False,
+    author='John Shepherd',
+    author_email='john@osrfoundation.org',
```
clalancette: Nit: We generally prefer the `@openrobotics.org` email addresses now.  We haven't been consistent on updating the old stuff, but for new stuff we should use that.

ament_clang_tidy/ament_clang_tidy/main.py
clalancette: https://github.com/ament/ament_lint/issues/157 is the new issue.

ament_cmake_clang_tidy/ament_cmake_clang_tidy-extras.cmake
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# copied from ament_cmake_clang_tidy/ament_cmake_clang_tidy-extras.cmake
+
+find_package(ament_cmake_test QUIET REQUIRED)
+
+include("${ament_cmake_clang_tidy_DIR}/ament_clang_tidy.cmake")
+
+ament_register_extension("ament_lint_auto" "ament_cmake_clang_tidy"
```
hidmic: @JShep1 @clalancette this is unfortunate as it may prevent us from backporting without causing havoc on packages that already rely on `ament_lint_auto` :(

hidmic: Is there a strong reason to have it hooked up? If not, and @dirk-thomas agrees, I'd say we remove it to preclude any issues once the patch release is out.

dirk-thomas: A package must be `find_package()`-ing this in order to trigger the extension being registered. So I don't see how this change would be a problem - even with `ament_lint_auto` since no package depends on this new package yet.

hidmic: > A package must be `find_package()`-ing this in order to trigger the extension being registered.

I see it commented out in `ament_lint_common`, yes. 

> So I don't see how this change would be a problem - even with ament_lint_auto since no package depends on this new package yet.

That's true of our packages. Could it ever be a problem if downstream packages start depending on it? If that's too far fetched, then I withdraw my concerns and suggestion.

dirk-thomas: If packages start depending on it they probably intend to use it?

hidmic: Argh, I wrote half of it. I meant to say that it may be a problem if downstream packages start depending on it while building core packages that depend on `ament_lint_auto` from source.

dirk-thomas: How would that effect downstream packages in any way?

hidmic: > A package must be find_package()-ing this in order to trigger the extension being registered.

If a downstream package depends on this linter, it'll get registered as an `ament_lint_auto` extension. If it gets registered as an `ament_lint_auto` extension, it'll get hooked up as a linter for packages that depend on `ament_lint_auto`. That is the case for core packages as well. Or at least that's my train of thought.

dirk-thomas: No package I am aware of declares a linter package as an exported dependency. They are usually only test dependencies.

hidmic: Alright, not a problem then :)

PR #163, opened 2019-07-26
Add error output, header out option, fix null error

### LINKED ISSUE #161: ament_clang_tidy doesn't show errors, thinks all headers are system headers
I'm trying to use the new `ament_clang_tidy`, but it's not (1) explaining the errors generated (2) emitting the 96357 warnings it thinks are in non-user code. I'm running Python 3.6.8 

```ament_clang_tidy src/ros2/demos```

> 1 error generated.
> Error while processing /opt/ros/master/src/ros2/demos/composition/include/composition/client_component.hpp.
> 2 errors generated.
> Error while processing /opt/ros/master/src/ros2/demos/composition/include/composition/listener_component.hpp.
> 3 errors generated.
> Error while processing /opt/ros/master/src/ros2/demos/composition/include/composition
> ...
> Error while processing /opt/ros/master/src/ros2/demos/quality_of_service_demo/rclcpp/src/liveliness.cpp.
> 95568 warnings and 71 errors generated.
> Error while processing /opt/ros/master/src/ros2/demos/quality_of_service_demo/rclcpp/src/utils.cpp.
> 96367 warnings and 72 errors generated.
> Error while processing /opt/ros/master/src/ros2/demos/quality_of_service_demo/rclcpp/src/utils.hpp.
> Suppressed 96357 warnings (96357 in non-user code).
> Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.

### PR DESCRIPTION
**Pull request by JShep1**
- Add error output to console.
- Add option for user to specify non-system header error output
- Fix bug with referencing `report` dictionary if no files were checked

Fixes #161 

Signed-off-by: John Shepherd <johnshepherd96@yahoo.com>

### DIFF
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -56,6 +56,13 @@ def main(argv=sys.argv[1:]):
         '--xunit-file',
         help='Generate a xunit compliant XML file')
     parser.add_argument(
+        '--header-filter',
+        help='Accepts a regex and displays errors from the specified non-system headers')
+    parser.add_argument(
+        '--system-headers',
+        action='store_true',
+        help='Displays errors from all system headers')
+    parser.add_argument(
         '--export-fixes',
         help='Generate a DAT file of recorded fixes')
     parser.add_argument(
@@ -71,10 +78,6 @@ def main(argv=sys.argv[1:]):
         action='store_true',
         help='Suppresses printing statistics about ignored warnings '
              'and warnings treated as errors')
-    parser.add_argument(
-        '--add-headers',
-        action='store_true',
-        help='Display errors from all non-system headers')
     args = parser.parse_args(argv)
 
     if not os.path.exists(args.config_file):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rotu** [comment]
I would add the `-quiet` option to `clang-tidy`. It does not suppress the output we're interested in but does suppress warnings about skipping system headers as well as the recommendation to pass the `-system-headers` argument (which `ament_clang_tidy` chockes on anyway).

**Comment by JShep1** [comment]
Sounds good, will add shortly, thanks for the recommendation

File: ament_clang_tidy/ament_clang_tidy/main.py [hunk 1]
```diff
@@ -66,6 +66,15 @@ def main(argv=sys.argv[1:]):
         '--fix-errors',
         action='store_true',
         help='Fix the suggested changes')
+    parser.add_argument(
+        '--quiet',
+        action='store_true',
+        help='Suppresses printing statistics about ignored warnings '
+             'and warnings treated as errors')
+    parser.add_argument(
+        '--add-headers',
```
jacobperron: Is there anything wrong with exposing the option as-is from clang-tidy? I.e. change this to `--header-filter` and accept a regex for header files.
If we do this, make sure to update the "help" description.

JShep1: Removed the binary option of specifying non-system header output and added support for ``-header-filter`` supplied with regex

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -55,6 +55,13 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--xunit-file',
         help='Generate a xunit compliant XML file')
+    parser.add_argument(
+        '--header-filter',
+        help='Accepts a regex and displays errors from the specified non-system headers')
+    parser.add_argument(
```
jacobperron: nitpick: I would re-order the arguments added by alphabetical order.

PR #165, opened 2019-07-30
Dashing backport for clang tidy and format linting

JShep1:
This PR serves as a backport fix for the addition of clang-tidy and custom config file support for clang-format.  These changes have previously been approved and merged into master.  See https://github.com/ament/ament_lint/pull/153, https://github.com/ament/ament_lint/pull/155, https://github.com/ament/ament_lint/pull/159

Signed-off-by: John Shepherd <john@openrobotics.org>

PR #166, opened 2019-07-30
Add python3-yaml depends to ament_clang_tidy/format

JShep1:
This is added due to a CI build failure (https://github.com/ament/ament_lint/pull/165) that occurs due to pyyaml being added to `setup.py` installation reqs (https://github.com/ament/ament_lint/pull/150) but python3-yaml not existing in `package.xml` as a dependency.

Signed-off-by: John Shepherd <john@openrobotics.org>


dirk-thomas:
Wouldn't an `exec_depend` be sufficient in both cases?

PR #167, opened 2019-07-30
Add missing yaml dep

rotu:
as per https://github.com/ament/ament_lint/pull/150#issuecomment-516569229


rotu:
Looks like @JShep1 beat me to it in https://github.com/ament/ament_lint/pull/166

PR #174, opened 2019-08-05
Fix indent after paren+newline

### LINKED ISSUE #173: Excessive indentation in wrapped `throw statement`
`ament_uncrustify` causes line continuations in throw statement to indent to 8 spaces instead of the 2 spaces recommended by the developer guide:

e.g.:
https://github.com/ros2/rclcpp/blob/9be3e08cd4d04193fb76cb6dd8f5db11d7199e2f/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L438-L441

```
  if (!result.successful) {
    throw rclcpp::exceptions::InvalidParameterValueException(
            "parameter '" + name + "' could not be set: " + result.reason);
  }
```

rotu:
Continuation indent after an open paren by one level instead of aligning to paren.

For some reason, this also prevents wrapping relative to the end of the `throw` keyword as well, even though there's not necessarily a paren there.

Fixes #173

A few examples from rclcpp:
before
```
  result_promise_.set_exception(std::make_exception_ptr(
      exceptions::UnawareGoalHandleError()));
//
    if (ret != RCL_RET_OK) {
      throw std::runtime_error(
              std::string("Couldn't initialize state machine for node ") +
              node_base_interface_->get_name());
//
  using FeedbackCallback =
    std::function<void (
        typename ClientGoalHandle<ActionT>::SharedPtr,
        const std::shared_ptr<const Feedback>)>;
```
after
```
  result_promise_.set_exception(std::make_exception_ptr(
    exceptions::UnawareGoalHandleError()));
//
    if (ret != RCL_RET_OK) {
      throw std::runtime_error(
        std::string("Couldn't initialize state machine for node ") +
        node_base_interface_->get_name());
    }
//
  using FeedbackCallback =
    std::function<void (
      typename ClientGoalHandle<ActionT>::SharedPtr,
      const std::shared_ptr<const Feedback>)>;
```

### DIFF
--- a/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
+++ b/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
@@ -954,7 +954,7 @@ indent_align_assign             = false    # false/true
 
 # Align continued statements at the '('. Default=True
 # If FALSE or the '(' is not followed by a newline, the next line indent is one tab.
-indent_align_paren              = true     # false/true
+indent_align_paren              = false     # false/true
 
 # Indent OC blocks at brace level instead of usual rules.
 indent_oc_block                 = false    # false/true


dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_linux/7800/)

dirk-thomas:
Changing this settings seems to have side effects when looking at what code `uncrustify` tries to reformat...

rotu:
> Changing this settings seems to have side effects when looking at what code `uncrustify` tries to reformat...

I looked at *everything* it did on the rclcpp codebase, and everything seemed correct. Do you have an example where it formats code incorrectly?

dirk-thomas:
If all the necessary changes are intended please create PRs to address all of these in order to pass CI with this change.

rotu:
Okay. But how to proceed? With #148, the changes can be made without causing CI to fail. This change, on the other hand, is *not* backwards compatible, so whether it's pushed to ament_uncrustify or consuming packages first, there *will* be CI failures. Is it possible to move this in and whitelist the resulting CI failures?

dirk-thomas:
> But how to proceed? With #148, the changes can be made without causing CI to fail.

I am not sure I understand how #148 is related to this change?

Before any change is merged it needs to be made sure that it passes CI without adding regressions. So if this change requires other code to be updated then there need to be PRs for that which can be tested together with this change. If we are convinced the update configuration matches out style guide better then we will be happy to merge the change - other code not covered by our CI will have to be updated for Eloquent accordingly.

> so whether it's pushed to ament_uncrustify or consuming packages first, there will be CI failures.

That is why all PRs need to be tested together and will be merged at the same time once CI passes.

> Is it possible to move this in and whitelist the resulting CI failures?

No, for that kind of invasive change it is not feasible to whitelist CI failures.

rotu:
I was contrasting to 148. Basically, I was wondering how to move this in concert with the other PRs necessary, since they'll all fail the pre-merge check. You partly answered my question - that we can put all the fixup PRs in CI at the same time and make sure that CI passes. They're still going to fail the pre-merge checks, however.

> No, for that kind of invasive change it is not feasible to whitelist CI failures.

How is this not feasible? I would expect that if our tools are broken (failing on code that agrees with the development standards and only allowing code that disagrees with it), this is exactly what whitelisting CI failures is for.

dirk-thomas:
> How is this not feasible? I would expect that if our tools are broken (failing on code that agrees with the development standards and only allowing code that disagrees with it), this is exactly what whitelisting CI failures is for.

dirk-thomas:
> How is this not feasible? I would expect that if our tools are broken (failing on code that agrees with the development standards and only allowing code that disagrees with it), this is exactly what whitelisting CI failures is for.

There is no desire to whitelist CI failures if you can resolve them right away. So if you want to modify the linter rules all existing code must be updated at the same time to comply.

dirk-thomas:
This change has undesired side effects for indentation in macros: see https://ci.ros2.org/job/ci_linux/9241/testReport/(root)/projectroot/uncrustify_2/

E.g. all these lines are unindented one level: https://github.com/ros2/rcl/blob/4b9c0a30beea07960cc9dd666dcf36ec1a18065f/rcl_action/src/rcl_action/action_client.c#L83-L87

rotu:
Curious. That appears to be an internal bug in uncrustify. I recommend either living with the bug or using `/* *INDENT-OFF* */` to locally disable Uncrustify around the code it malfunctions on.

dirk-thomas:
The two recent commits seems to be a no-op together. Please summarize the current state, the effect it has, and your proposal how to proceed if you would like to be reviewed beyond the already made comments.

rotu:
Yep. Those two recent commits were me misunderstanding a configuration option. I just now force pushed to erase those commits.

The "undesired side effects for indentation in macros" here are an upstream bug in uncrustify that this PR surfaces. I recommend working around this bug by either living with the bad formatting in macros or manually formatting this and suppressing uncrustify with `/* *INDENT-OFF* */`

dirk-thomas:
How much code is affected by this change? Can you show examples of the proposed changes in order to comply with the change? Are you planning on making PRs for the necessary code changes?

rotu:
No code is "affected" by this change. It makes it so CI alerts on previously missed violations of the documented code style.

I'm planning on making some PRs for the necessary code changes, though the full scope is beyond my capacity.

rotu:
@dirk-thomas what more information is needed?

dirk-thomas:
> what more information is needed?

As described before: https://github.com/ament/ament_lint/pull/174#issuecomment-611158357

To be explicit:
* There need to be PRs for all code changes necessary to satisfy the changes linter configuration. We simply can't a linter change if CI has linter warnings afterwards.
* All those changes need to be desired (or if not at least accepted if they pose a regression). And there have been cases like this being brought up in the one PR updating code).

rotu:
Okay. I'll let someone else fix https://github.com/ament/ament_lint/issues/173 then. It seems it's easier to leave this broken then jump through so many hoops.

PR #176, opened 2019-08-09
Ability to exclude folders with ament_pep257

deitry:
I discovered that ament_pep257 exclude option accepts only specific file names but not folder names.

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -107,6 +107,11 @@ def main(argv=sys.argv[1:]):
     return rc
 
 
+def _filename_in_excludes(filename, excludes):
+    absname = os.path.abspath(filename)
+    return any(os.path.commonpath([absname, e]) == e for e in excludes)
+
+
 def generate_pep257_report(paths, excludes, ignore):
     conf = ConfigurationParser()
     sys_argv = sys.argv
@@ -126,7 +131,7 @@ def generate_pep257_report(paths, excludes, ignore):
     files_dict = {}
     if LooseVersion(pydocstyle.__version__) >= LooseVersion('2.0.0'):
         for filename, checked_codes, ignore_decorators in files_to_check:
-            if os.path.abspath(filename) in excludes:
+            if _filename_in_excludes(filename, excludes):
                 continue
             files_dict[filename] = {
                 'select': checked_codes,
@@ -134,7 +139,7 @@ def generate_pep257_report(paths, excludes, ignore):
             }
     else:
         for filename, select in files_to_check:
-            if os.path.abspath(filename) in excludes:
+            if _filename_in_excludes(filename, excludes):
                 continue
             files_dict[filename] = {
                 'select': select,


ament_pep257/ament_pep257/main.py
```diff
@@ -107,6 +107,13 @@ def main(argv=sys.argv[1:]):
     return rc
 
 
+def _filename_in_excludes(filename, excludes):
```
christophebedard: Just a suggestion: something a bit more Pythonic could be

```python
def _filename_in_excludes(filename, excludes):
    return any([os.path.abspath(filename).startswith(e) for e in excludes])
```

:smiley:

deitry: Agree, thank you

deitry: In fact, since it is one-liner, there is no need in function at all. It may be inlined in the places of use

deitry: Amended.

> In fact, since it is one-liner, there is no need in function at all. It may be inlined in the places of use
Though I decided to keep function instead of copy-paste.

christophebedard: My bad, this is actually worse than your original function! [Using a generator expression](https://stackoverflow.com/a/16505590/6476709) (without the `[]`) is better than using a list comprehension.

ament_pep257/ament_pep257/main.py
```diff
@@ -107,6 +107,10 @@ def main(argv=sys.argv[1:]):
...
+    return any([os.path.abspath(filename).startswith(e) for e in excludes])
```
dirk-thomas: You can't use `startswith` like this to check if the `filename` is in a list of directories.

Counter example: `exclude: /foo`, `filename: /foobar` which would incorrectly return `True`.

deitry: Good point, thank you.
Now checking if `filename` is exact `exclude` or is prefixed with `exclude/`.

dirk-thomas: This won't work in Windows which uses backslashes.

deitry: Thanks, completely forgot about that.
The newest version got rid of `startwith` in favor of `commonpath`.

dirk-thomas: `commonpath` will have the same problem. The easiest way to do the check is something like this:

```
path = pathlib.Path(filename)
exclude = pathlib.Path(exclude)
exclude == p or exclude in path.parents
```

deitry: > `commonpath` will have the same problem

Tested on Windows 10, Python 3.6.8:
```python
>>> os.path.abspath('1.txt')
'C:\\test\\1.txt'
>>> f = "C:\\test\\1.txt"
>>> f1 = "C:\\test2\\1.txt" 
>>> e = "C:\\test"
>>> os.path.commonpath([f,e]) == e
True
>>> os.path.commonpath([f1,e]) == e 
False
```

Am I missed something?

dirk-thomas: Oh, I confused it with `commonprefix`. `commonpath` is good as is.

dirk-thomas:
Thank you for the patch.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7859/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3922/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6413/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7720/)

The test failures across all platforms are unrelated to this change and already present on master.

PR #178, opened 2019-08-10
re-add path insertion removed from wrong package

Arnatious:
fixes removal of path insertion from flake8


dirk-thomas:
Thanks for the quick fix.

PR #181, opened 2019-08-15
fix cmake plugin to use correct config file argument keyword

Arnatious:
ament_mypy.cmake used wrong flag for passing config files to ament_mypy. Replaced with correct flag. 


dirk-thomas:
Thanks for the fix.

PR #184, opened 2019-08-23
Prepare 0.7.8

hidmic:


dirk-thomas:
`ament_clang_tidy/setup.py` isn't being updated / has a different version number.

hidmic:
Good catch ! Force pushed to keep a single commit to FF.

hidmic:
Alright, going in.

PR #191, opened 2019-09-11
Add pytest marker for mypy

emersonknapp:
Signed-off-by: Emerson Knapp <eknapp@amazon.com>


dirk-thomas:
Thanks for the patch.

PR #198, opened 2019-10-22
Behave better in multi-project cmake

rotu:
If a sub-project uses ament_lint, don't apply it to the top-level project headers


dirk-thomas:
Thanks for the patch.

PR #199, opened 2019-10-23
fix doc for ament_cmake_copyright

rotu:
fix function name


dirk-thomas:
Thanks for the improvement. Please use a more specific title next time. Also please sign off the commit to pass the DCO check.

dirk-thomas:
@rotu Please comment on the ticket in the future when making changes otherwise there is no notification and we might not revisit the ticket (as it happened for this one for almost a month).

@wjwwood Thanks for catching this.

PR #201, opened 2019-12-09
Ignore pydocstyle D401.

nuclearsandwich:
pydocstyle 5.0.0 fixed this check which many of our docstrings are now
failing. There is an ongoing discussion about whether or not we want to
follow it.

This is a straw-man to disable it assuming the answer is not yes.

Closes https://github.com/ros2/build_cop/issues/255


nuclearsandwich:
Closing in favor of resolving the errors in all source packages.

PR #204, opened 2020-01-29
Changing uncrustify config option which fixes nested templates.

JWhitleyAStuff:
Please see the originally-reported issue on `uncrustify` upstream here: 

https://github.com/uncrustify/uncrustify/issues/2613 and the source of the issue: 

https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/issues/235

According to the documentation, this can have unintended consequences:

```
# Allow interpreting '>=' and '>>=' as part of a template in 'void f(list<list<B>>=val);'.
# If True, 'assert(x<0 && y>=3)' will be broken. Default=False
# Improvements to template detection may make this option obsolete.
```

So it is definitely worth some discussion and testing.

### DIFF
--- a/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
+++ b/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg
@@ -25,7 +25,7 @@ string_replace_tab_chars        = false    # false/true
 # Allow interpreting '>=' and '>>=' as part of a template in 'void f(list<list<B>>=val);'.
 # If True, 'assert(x<0 && y>=3)' will be broken. Default=False
 # Improvements to template detection may make this option obsolete.
-tok_split_gte                   = false    # false/true
+tok_split_gte                   = true     # false/true
 
 # Override the default ' *INDENT-OFF*' in comments for disabling processing of part of the file.
 disable_processing_cmt          = " *INDENT-OFF*"      # string


dirk-thomas:
CI build running `uncrustify` tests only: [Build Status](https://ci.ros2.org/job/ci_linux/9171/)

dirk-thomas:
> # If True, 'assert(x<0 && y>=3)' will be broken. Default=False

Since our code style prescribes spaces around the operator I don't think we will run into this case for code passing the linter.

dirk-thomas:
The CI builds shows an undesired side effect: https://ci.ros2.org/job/ci_linux/9176/testReport/rviz_default_plugins/uncrustify/src_rviz_default_plugins_displays_robot_model_robot_model_display_cpp/

JWhitleyAStuff:
@dirk-thomas - Interesting. Yeah, I figured it would run into something. Thoughts? Should I try to maybe improve the template detection code in `uncrustify` as mentioned in the comment? Whether this feature is enabled or disabled, I would say there is a bug in `uncrustify` that causes an issue either way - just that the code in the CI builds for the demos and such don't use the syntax that originally triggered the bug for us.

dirk-thomas:
Iterating on improving `uncrustify` is always a good idea. Otherwise / until then we just have to work around its limitations in our code / config. Not sure which change is easier.

JWhitleyAStuff:
Our use-case of having a template-within-a-template is probably less common. I'll re-open the ticket on the `uncrustify` side until the issue is fixed. Thanks for checking into it!

PR #206, opened 2020-01-29
Fix bug, allows using ament_copyright with bsd3

Blast545:
To reproduce this bug you have to use the ament_copyright tool to try adding missing headers to a file, using the bsd2 license. After sourcing the ROS workspace:

`ament_copyright --add-missing 'osrf' bsd2 src/ament/ament_lint/ament_copyright/test/cases/bsd_license/case.py`

This will cause
`KeyError: 'copyright_holder'`

Format of the bsd2 license expected a command to pass a 'copyright_holder' key as a parameter to include in the license. However, the current CLI does not includes this. This quick fix uses as copyright holder this: 'the copyright holder', so the text match the condition as expressed in: https://opensource.org/licenses/BSD-3-Clause

Later on it could be discussed if it's a desirable solution to include adding this copyright holder variable in the cli.

### DIFF
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -218,7 +218,9 @@ def main(argv=sys.argv[1:]):
 
 def add_missing_header(file_descriptors, name, license_, verbose):
     copyright_ = 'Copyright %d %s' % (int(time.strftime('%Y')) - 1 + 1, name)
-    header = license_.file_header.format(**{'copyright': copyright_})
+    header = license_.file_header.format(**{
+        'copyright': copyright_,
+        'copyright_holder': 'the copyright holder'})
     lines = header.splitlines()
 
     if verbose:
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -217,7 +217,7 @@ def main(argv=sys.argv[1:]):
 
 
 def add_missing_header(file_descriptors, name, license_, verbose):
-    copyright_ = 'Copyright %d, %s' % (int(time.strftime('%Y')) - 1 + 1, name)
+    copyright_ = 'Copyright %d %s' % (int(time.strftime('%Y')) - 1 + 1, name)
     header = license_.file_header.format(**{
         'copyright': copyright_,
         'copyright_holder': name})
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -217,10 +217,10 @@ def main(argv=sys.argv[1:]):
 
 
 def add_missing_header(file_descriptors, name, license_, verbose):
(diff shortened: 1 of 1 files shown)


ament_copyright/ament_copyright/main.py
```diff
@@ -218,7 +218,9 @@ def main(argv=sys.argv[1:]):
 
 def add_missing_header(file_descriptors, name, license_, verbose):
     copyright_ = 'Copyright %d %s' % (int(time.strftime('%Y')) - 1 + 1, name)
-    header = license_.file_header.format(**{'copyright': copyright_})
+    header = license_.file_header.format(**{
+        'copyright': copyright_,
+        'copyright_holder': 'the copyright holder'})
```
dirk-thomas: "the copyright holder" doesn't seem to be a reasonable replacement. Either it should be an actual name or the template should use that literal string instead of a placeholder.

Blast545: Was thinking a little about it just now, and actually the solution is to use the 'name' variable passed as argument. This way the generated copyright might look as this one:
[class_loader.cpp](https://github.com/ros/class_loader/blob/melodic-devel/src/class_loader.cpp)

Where the copyright in the top has the copyright year autogenerated, and the passed name is used in both places

ament_copyright/ament_copyright/main.py
```diff
@@ -217,8 +217,10 @@ def main(argv=sys.argv[1:]):
 
 
 def add_missing_header(file_descriptors, name, license_, verbose):
-    copyright_ = 'Copyright %d %s' % (int(time.strftime('%Y')) - 1 + 1, name)
-    header = license_.file_header.format(**{'copyright': copyright_})
+    copyright_ = 'Copyright %d, %s' % (int(time.strftime('%Y')) - 1 + 1, name)
```
dirk-thomas: The majority of existing code doesn't seem to have a comma between the year and the name. Why should that be changed in the template?

Blast545: I was thinking in some new licenses notices I have seen with the comma, I will remove it to avoid changing the standard.

Blast545:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9183/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4888/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7490/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9093/)

dirk-thomas:
Based on https://answers.ros.org/question/353760/ament_copyright-add-missing-gives-keyerror-copyright_holder/ this should probably be backported to Eloquent. @mjcarroll Can you please do that (and check if it is also needed for Dashing)?

Issue #149, opened 2019-07-09
ament flake8 does not ignore packages with AMENT/COLCON_IGNORE


jpsamper2009:
Other `ament` linters will automatically filter out files from directories that have an AMENT_IGNORE: e.g.

-  ament_copyright: https://github.com/ament/ament_lint/blob/master/ament_copyright/ament_copyright/crawler.py#L28
- ament_cpplint: https://github.com/ament/ament_lint/blob/master/ament_cpplint/ament_cpplint/main.py#L208
- ament_lint_cmake:  https://github.com/ament/ament_lint/blob/master/ament_lint_cmake/ament_lint_cmake/main.py#L128

Ideally, `ament flake8` would be consistent with this behavior.

dirk-thomas:
Please consider to provide a pull request for this enhancement.

PR #211, opened 2020-02-05
Exclude folders having AMENT_IGNORE in ament_flake8

left4taco:
After this PR, `ament_flake8` will walk through the package folder and exclude any directories that have AMENT_IGNORE or COLCON_IGNORE. 

Closes https://github.com/ament/ament_lint/issues/149

### DIFF
--- a/ament_flake8/ament_flake8/main.py
+++ b/ament_flake8/ament_flake8/main.py
@@ -70,6 +70,14 @@ def main(argv=sys.argv[1:]):
         print("Could not find config file '%s'" % args.config_file, file=sys.stderr)
         return 1
 
+    # Filter out folders having AMENT_IGNORE
+    if args.excludes is None:
+        args.excludes = []
+    for dirpath, dirnames, filenames in os.walk(os.getcwd()):
+        if 'AMENT_IGNORE' in filenames:
+            dirnames[:] = []
+            args.excludes.append(dirpath)
+
     report = generate_flake8_report(
         args.config_file, args.paths, args.excludes,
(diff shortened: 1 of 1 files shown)


left4taco:
@dirk-thomas Could you point me to the right person to review this PR? Thanks.

dirk-thomas:
Anyone from @ament/team can review this PR.

ament_flake8/ament_flake8/main.py
```diff
@@ -70,6 +70,14 @@ def main(argv=sys.argv[1:]):
         print("Could not find config file '%s'" % args.config_file, file=sys.stderr)
         return 1
 
+    # Filter out folders having AMENT_IGNORE or COLCON_IGNORE
+    if args.excludes is None:
+        args.excludes = []
+    for dirpath, dirnames, filenames in os.walk(os.getcwd()):
+        if 'AMENT_IGNORE' in filenames or 'COLCON_IGNORE' in filenames:
```
jacobperron: I don't think it makes sense to ignore directories with `COLCON_IGNORE` since ament is colcon agnostic.

The other linters similarly only check for `AMENT_IGNORE`.

wjwwood: I agree, you should just use `AMENT_IGNORE` instead of `COLCON_IGNORE`.

left4taco: @jacobperron @wjwwood Thanks for pointing it out. I've removed the `COLCON_IGNORE`. Please review again.

PR #213, opened 2020-02-18
Fix ament_clang_tidy

mm318:
### Changes made
- The <path> argument is now meant to point towards a compile_commands.json file or a directory that contains it, instead of source files
- Unit tests source files are currently skipped because gtest has linter violations that the ROS2 package developer wouldn't have control over

### How to use/test
Follow the README to install necessary dependencies (for example on Ubuntu 18.04, do `apt install clang-tidy-6.0 clang-tools-6.0 python-yaml`).
In a workspace that contains the packages that you want to lint (or want to test the linter on):
- Build/rebuild with `colcon build --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
- Run `ament_clang_tidy build/`
    - Expect to see a list of linter warnings/errors printed to `stdout`
    - Prior to this fix, all of the linter errors are just complaints about missing header files
- Run `ament_clang_tidy build/ --xunit-file test.xml`
    - Expect to see a list of linter warnings/errors printed to `stdout`
    - Expect to see the same list of warnings/errors in the outputted `test.xml` file

### DIFF
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -14,9 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import defaultdict
 import argparse
+from collections import defaultdict
 import copy
+import json
 import multiprocessing.pool
 import os
 import re
@@ -24,11 +25,9 @@ import subprocess
 import sys
 import time
 
-from xml.sax.saxutils import escape
 from xml.sax.saxutils import quoteattr
 
 import yaml
-import json
 
 
 def main(argv=sys.argv[1:]):
@@ -56,7 +55,7 @@ def main(argv=sys.argv[1:]):
         '--jobs',
         type=int,
         default=1,
-        help="number of clang-tidy jobs to run in parallel")
+        help='number of clang-tidy jobs to run in parallel')
 
     # not using a file handle directly
     # in order to prevent leaving an empty file when something fails early
@@ -108,7 +107,7 @@ def main(argv=sys.argv[1:]):
     for file in files:
         package_dir = os.path.dirname(file)
         package_name = os.path.basename(package_dir)
-        print("linting " + package_name + "...")
+        print('linting ' + package_name + '...')
         async_outputs.append(pool.apply_async(invoke_clang_tidy, (clang_tidy_bin, file, args)))
     pool.close()
     pool.join()
@@ -226,11 +225,11 @@ def invoke_clang_tidy(clang_tidy_bin, compilation_db_path, args):
     for item in db:
         # exclude gtest sources from being checked by clang-tidy
         if is_gtest_source(os.path.basename(item['file'])):
-          continue
+            continue
         # exclude unit test sources from being checked by clang-tidy
         # because gtest macros are problematic
         if is_unittest_source(package_name, item['file']):
-          continue
+            continue
 
         full_cmd = cmd + [item['file']]
         # print(' '.join(full_cmd))
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -67,6 +67,13 @@ def main(argv=sys.argv[1:]):
         '--export-fixes',
         help='Generate a DAT file of recorded fixes')
     parser.add_argument(
+        '--fix-errors',
+        action='store_true',
+        help='Fix the suggested changes')
+    parser.add_argument(
+        '--header-filter',
+        help='Accepts a regex and displays errors from the specified non-system headers')
+    parser.add_argument(
         '--quiet',
         action='store_true',
         help='Suppresses printing statistics about ignored warnings '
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Maybe you can elaborate why removing existing options like `--fix-errors` or `--header-filter` is necessary to address the question about how to use this linter?

mm318:
`--fix-errors` is not implemented in this PR. I was thinking of adding it in a separate PR. By the way, `clang-tidy` can't automatically fix all the issues that it flags.

The way the code in this PR sets up `--header-filter` by default works, and I think if the user were to override it, the user would either get too many irrelevant linter warnings/errors or too few relevant ones. I could put this option back, though I would then recommend against using it.

mm318:
Hi @dirk-thomas, can you please take a look again?

dirk-thomas:
The current patch does multiple unrelated changes (e.g. change semantic of the `paths` argument, use multiprocessing, unrelated style changes, change invoked scripts and arguments, etc.), which make it more complex and unnecessary complicated to review. Please split the contribution into separate PRs each focus on a single change.

claireyywang:
@mm318 friendly ping

mm318:
I will get to it before end of next week. Thanks.

zmichaels11:
@mm318 any status updates on this?

thomas-moulard:
@mm318 any status updates on this?

mm318:
I have updated the pull request by reducing the amount of changes.

claireyywang:
@mm318 Thanks for contributing! Can you plz include more information on the PR description about the final list of changes this PR introduces, how to test and what to expect from testing?

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -88,13 +89,13 @@ def main(argv=sys.argv[1:]):
...
-    if not files:
-        print('No files found', file=sys.stderr)
+    compilation_dbs = get_compilation_db_files(args.paths)
+    if not compilation_dbs:
+        print('No compilation database files found', file=sys.stderr)
         return 1
 
     bin_names = [
-        'clang-tidy',
+        # 'clang-tidy',
```
mm318: Only use `clang-tidy` 6.0 for now, because it is the most ubiquitous, the default version of `clang-tidy` on Ubuntu Bionic seems to be a moving target, and the parsing of linter warning/error messages has very recently been made difficult by an update to the default version of `clang-tidy` on Ubuntu Bionic. This is also consistent with instructions in the README that's part of this pull request.

mm318:
Hi @claireyywang, I have updated the pull request description with more information. Thanks.

ament_clang_tidy/ament_clang_tidy/main.py
dirk-thomas: The master branch is not targeting Ubuntu Bionic anymore but Ubuntu Focal. And the default version of `clang-tidy` in Focal is 10.0. So I am not sure it makes sense to target version 6 here...

mm318: Will the `master` branch be backported to the branches that target Ubuntu Bionic, for example for the Dashing distro? Since the `clang-tidy-6.0` apt package is valid for both Ubuntu Bionic and Focal, I think we should stick with `clang-tidy` 6.0 for now, then maybe backport, then target `clang-tidy` 10.0 (along with other features that I have in mind for this package).

dirk-thomas: Changes need to be explicitly backport through PRs. It won't happen automatically.

Using version 10 on Focal should happen before the upcoming freeze date. I don't mind which way that happens (target version 10 in this PR and then make a backport PR for Eloquent / Dashing which uses version 6) orland this PR targeting 6 and followingup with a PR to switch to version 10 with the intention to only backport the first.

mm318: Let's land this PR targeting `clang-tidy` 6.0, with a follow-up PR to switch to `clang-tidy` 10.0.

mm318:
Hi @claireyywang,

> Got a `KeyError` when testing on [`demos` repo](https://github.com/ros2/demos/)

I have fixed the issue that you found.

> Also the process took more than 5 mins to output all the warnings, is that expected?

Yes. It takes this long because the `demos` repo has 14 packages and I have removed the multiprocessing feature from this pull request to reduce the scope of changes. Also, `clang-tidy` is a more complex and compute intensive linter than the other linters.

Thanks!

claireyywang:
CI 
Linux [Build Status](https://ci.ros2.org/job/ci_linux/9887/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5505/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/8045/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/9744/)

mm318:
Thanks for merging!

claireyywang:
> Thanks for merging!

np!

PR #214, opened 2020-02-18
pass explicit language to uncrustify

dirk-thomas:
Since the latest version of `uncrustify` fails for C files otherwise and upstream will unlikely fix the regression: see uncrustify/uncrustify#2641.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9343/) (unrelated test failures)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5026/) (same)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7650/) (same)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9291/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/141/) (unrelated compiler warnings)

### DIFF
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -64,6 +64,11 @@ def main(argv=sys.argv[1:]):
         default=[],
         help='Exclude specific file names and directory names from the check')
     parser.add_argument(
+        '--language',
+        choices=['C', 'C++'],
+        help="Passed to uncrustify as '-l <language>' to force a specific "
+             'rather then choose based on the file extensions')
+    parser.add_argument(
         '--reformat',
         action='store_true',
         help='Reformat the files in place')
@@ -101,7 +106,7 @@ def main(argv=sys.argv[1:]):
 
         files_by_language = get_files(
             args.paths, {'C': c_extensions, 'C++': cpp_extensions},
-            args.exclude)
+            excludes=args.exclude, language=args.language)
         if not files_by_language:
             print('No files found', file=sys.stderr)
             return 1
@@ -202,10 +207,10 @@ def find_executable(file_name, additional_paths=None):
     return shutil.which(file_name, path=path)
 
 
-def get_files(paths, extension_types, excludes=[]):
+def get_files(paths, extension_types, excludes=[], language=None):
     extensions_with_dot_to_language = {
-        f'.{extension}': language
-        for language, extensions in extension_types.items()
+        f'.{extension}': language or ext_language
+        for ext_language, extensions in extension_types.items()
         for extension in extensions
     }
     files = defaultdict(list)
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -260,10 +260,10 @@ def invoke_uncrustify(
         input_files = [os.path.relpath(f, start=cwd) for f in input_files]
 
     try:
-        cmd = [
-            uncrustify_bin,
-            '-c', args.config_file,
-            '-l', language,
+        cmd = [uncrustify_bin, '-c', args.config_file]
+        if language:
+            cmd += ['-l', language]
+        cmd += [
             '--prefix', temp_path,
             '--suffix', suffix]
         cmd.extend(input_files)
@@ -321,7 +321,10 @@ def invoke_uncrustify(
         # reinvoke uncrustify for previously changed files
         input_files = changed_files
         try:
-            cmd = [uncrustify_bin, '-c', args.config_file, '--suffix', suffix]
+            cmd = [uncrustify_bin, '-c', args.config_file]
+            if language:
+                cmd += ['-l', language]
+            cmd += ['--suffix', suffix]
             cmd.extend(input_files)
             subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as e:
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -67,7 +67,7 @@ def main(argv=sys.argv[1:]):
         '--language',
         choices=['C', 'C++'],
         help="Passed to uncrustify as '-l <language>' to force a specific "
-             'rather then choose based on the file extensions')
+             'language rather then choosing one based on file extension')
     parser.add_argument(
         '--reformat',
         action='store_true',
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -119,7 +119,6 @@ def main(argv=sys.argv[1:]):
         all_input_files = []
         all_output_files = []
         for language, input_files in files_by_language.items():
-            input_files = files_by_language[language]
             all_input_files += input_files
             output_files = invoke_uncrustify(
                 uncrustify_bin, input_files, args, language, temp_path, suffix)
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import argparse
+from collections import defaultdict
 from configparser import ConfigParser
 import difflib
 import filecmp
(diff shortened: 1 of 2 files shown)


ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -106,106 +111,22 @@ def main(argv=sys.argv[1:]):
...
-            uncrustified_files = [f + suffix for f in input_files]
-            i += 1
-            if i >= 5:
-                print("'uncrustify' did not settle on a final result even "
-                      "after %d invocations" % i, file=sys.stderr)
-                return 1
+        all_input_files = []
+        all_output_files = []
+        for language, input_files in files_by_language.items():
+            input_files = files_by_language[language]
```
hidmic: @dirk-thomas unnecessary?

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -106,106 +111,22 @@ def main(argv=sys.argv[1:]):
...
+            all_input_files += input_files
+            output_files = invoke_uncrustify(
+                uncrustify_bin, input_files, args, language, temp_path, suffix)
+            all_output_files += output_files
 
         # compute diff
-        for index, filename in enumerate(files):
-            modified_filename = output_files[index]
+        for filename, modified_filename in sorted(zip(all_input_files, all_output_files)):
```
hidmic: @dirk-thomas does `invoke_uncrustify` ensure `input_files` order is maintained in `output_files`?

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -282,8 +203,12 @@ def find_executable(file_name, additional_paths=None):
     return shutil.which(file_name, path=path)
 
 
-def get_files(paths, extensions, excludes=[]):
-    files = []
+def get_files(paths, extension_types, excludes=[]):
+    extensions_with_dot_to_language = {}
+    for language, extensions in extension_types.items():
+        for extension in extensions:
+            extensions_with_dot_to_language['.%s' % extension] = language
```
hidmic: @dirk-thomas nit^2: consider turning this into a dictionary comprehension.

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -301,12 +226,118 @@ def get_files(paths, extensions, excludes=[]):
...
+                    language = extensions_with_dot_to_language.get(ext, None)
+                    if language is not None:
+                        files[language].append(
+                            os.path.normpath(os.path.join(dirpath, filename)))
         if os.path.isfile(path):
-            files.append(path)
-    return [os.path.normpath(f) for f in files]
+            _, ext = os.path.splitext(path)
+            language = extensions_with_dot_to_language.get(ext, None)
+            files[language].append(os.path.normpath(path))
```
hidmic: @dirk-thomas consider checking if `language is not None` here too, or drop the default value in `get()` above.

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -301,12 +226,118 @@ def get_files(paths, extensions, excludes=[]):
...
+                    # on Windows os.rename doesn't replace an existing dst
+                    os.remove(dst_filename)
+                os.rename(output_filename, dst_filename)
+                changed_files.append(dst_filename)
+        if not changed_files:
+            break
+        # reinvoke uncrustify for previously changed files
+        input_files = changed_files
+        try:
+            cmd = [uncrustify_bin, '-c', args.config_file, '--suffix', suffix]
```
hidmic: @dirk-thomas uh, isn't the language flag necessary here as well?

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Can you please elaborate what you mean.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Yes, the `output_files` list is computed by a list comprehension over the `input_files`.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Not sure if the double for dict comprehension is easier to read for most people but sure: f82b7dd95bf94bc82864952b41cad426fe27b380.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Why should this check for `language is not None`? We need the `None` key to enumerate all files with unknown extension.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Absolutely, good catch: b3f4a448c19444ae98103bfba25e1aa7470ed93c

And since language can be `None` it needs to be added conditionally.

ament_uncrustify/ament_uncrustify/main.py
hidmic: There's no need to retrieve `input_files` again from the dictionary that is being iterated.

ament_uncrustify/ament_uncrustify/main.py
hidmic: Oh, ok, to deal with the explicitly given files with unknown extension. Understood. Disregard.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Ah, now I got it: removed in c330e7d8c834a83f87ff69d59d93661433cd0c38.

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -282,8 +202,13 @@ def find_executable(file_name, additional_paths=None):
     return shutil.which(file_name, path=path)
 
 
-def get_files(paths, extensions, excludes=[]):
-    files = []
+def get_files(paths, extension_types, excludes=[]):
+    extensions_with_dot_to_language = {
+        (language, extension)
```
hidmic: @dirk-thomas hmm I think this is a set comprehension, shouldn't it be `language: extension` instead?

dirk-thomas: Indeed, another fix: 3c76a9e83c9f09ac2580b057718020a8e8313fc5.

dirk-thomas:
Another CI build to double check the recent iterations - Linux only: [Build Status](https://ci.ros2.org/job/ci_linux/9381/)

dirk-thomas:
For `urdf` an option to pass an explicit language was necessary (same as already done for `cppcheck`) since the package uses `.h` for C++ files.

* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/9394/)

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -54,12 +56,18 @@ def main(argv=sys.argv[1:]):
...
+                 "'.%s'" % e for e in sorted(c_extensions + cpp_extensions)]))
     parser.add_argument(
         '--exclude',
         nargs='*',
         default=[],
         help='Exclude specific file names and directory names from the check')
+    parser.add_argument(
+        '--language',
+        choices=['C', 'C++'],
+        help="Passed to uncrustify as '-l <language>' to force a specific "
```
hidmic: @dirk-thomas nit: 
```suggestion
        help="Passed to uncrustify as '-l <language>' to force a specific language"
```

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -54,12 +56,18 @@ def main(argv=sys.argv[1:]):
...
+             'rather then choose based on the file extensions')
```
hidmic: @dirk-thomas nit: 
```suggestion
             'rather than choosing one based on file extension')
```

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Thanks, done in 7f015e5394989595167130a71c649eface660017 to balance the line length.

ament_uncrustify/ament_uncrustify/main.py
dirk-thomas: Thanks, done in 7f015e5394989595167130a71c649eface660017 to balance the line length.

PR #219, opened 2020-03-10
Suppress internalAstError

rotu:
`cppcheck` can, in some cases, construct an invalid parse tree.
http://build.ros2.org/user/rotu/my-views/view/CycloneDDS/job/Fci__nightly-cyclonedds_ubuntu_focal_amd64/lastCompletedBuild/testReport/rclcpp/cppcheck/error__internalAstError__src_rclcpp_clock_cpp_159_/

```
- rclcpp.cppcheck error: internalAstError (src/rclcpp/clock.cpp:159)
  <<< failure message
    Syntax Error: AST broken, 'if' doesn't have two operands.
  >>>
```
This error sounds like a syntax error, but is in fact, not. (cppcheck doesn't seem to even find the syntax error in `if(1,1){}`)
This commit causes such errors to be silently ignored, as they are not actionable.

### DIFF
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -134,7 +134,8 @@ def main(argv=sys.argv[1:]):
            '-q',
            '-rp',
            '--xml',
-           '--xml-version=2']
+           '--xml-version=2',
+           '--suppress=internalAstError']
     if args.language:
         cmd.extend(['--language={0}'.format(args.language)])
     for include_dir in (args.include_dirs or []):


ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -126,15 +126,15 @@ def main(argv=sys.argv[1:]):
                 return 0
 
             return 188
-
```
dirk-thomas: Nitpick: please revert unrelated changes.

rotu: oops!

rotu:
> Wouldn't it be better to mark the specific file as "skipped" to not pretend it was checked when it actually wasn't?

It does continue to find errors, just so long as they don't depend on that part of the AST. Adding `int x=1;` both inside and outside the mis-parsed `if` statement flags the missing space in both places.

dirk-thomas:
Thanks for details. Sounds reasonable then.

dirk-thomas:
(Will merge without CI since ci.ros2.org has issues at the moment and this option is available in `cppcheck` on Bionic so I don't expect any regressions.)

PR #220, opened 2020-03-11
Generate xunit files valid for the junit10.xsd

j-rivero:
With the new xunit plugin version 2.x the xunit files generated should be compliant with the [junit10.xsd](https://raw.githubusercontent.com/jenkinsci/xunit-plugin/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd).

The current code in different packages writing xunit files lacks the required attribute `errors` and use `skipped` instead of `skip`. For the `skip` attribute the change is trivial. For the `errors` attribute the description is: 
> The total number of tests in the suite that errored. An errored test is one that had an unanticipated problem. e.g., an unchecked throwable; or a problem with the implementation of the test

The code of linters in this repository exit the test to the system if it founds some problems with the manipulation of the different tools (like the lack of cppcheck in ament_cppcheck) so I'm assuming here that if the test is not aborted before the number of these `errors` is 0.

Locally the files can be validated using (download the xsd from the link above):
`xmllint --schema junit-10.xsd path/to/ppcheck.xunit.xml --noout`

Waiting the buildfarm test for this job: [Build Status](https://ci.ros2.org/job/ci_linux/9598/)

### DIFF
--- a/ament_clang_format/ament_clang_format/main.py
+++ b/ament_clang_format/ament_clang_format/main.py
@@ -303,6 +303,7 @@ def get_xunit_content(report, testname, elapsed):
 <testsuite
   name="%(testname)s"
   tests="%(test_count)d"
+  errors="0"
   failures="%(error_count)d"
   time="%(time)s"
 >
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -247,6 +247,7 @@ def get_xunit_content(report, testname, elapsed):
 <testsuite
   name="%(testname)s"
   tests="%(test_count)d"
+  errors="0"
   failures="%(error_count)d"
   time="%(time)s"
 >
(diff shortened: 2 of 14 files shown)


dirk-thomas:
> I'm assuming here that if the test is not aborted before the number of these errors is `0`.

Should this be updated then to signal an `errors="1"` instead of `failures="1"` then? https://github.com/ament/ament_cmake/blob/37ed0dcde921f4f0dd29e0e6cf64c55f85e193f0/ament_cmake_test/ament_cmake_test/__init__.py#L307

j-rivero:
> > I'm assuming here that if the test is not aborted before the number of these errors is `0`.
> 
> Should this be updated then to signal an `errors="1"` instead of `failures="1"` then? https://github.com/ament/ament_cmake/blob/37ed0dcde921f4f0dd29e0e6cf64c55f85e193f0/ament_cmake_test/ament_cmake_test/__init__.py#L307

In the case of detecting segfaults or other problems not related to bad results of the test I think so. Let's follow this here: https://github.com/ament/ament_cmake/pull/225

PR #221, opened 2020-03-19
add new main_with_errors() API that also returns the error strings if any

dirk-thomas:
Instead of returning an integer return code this patch lets the `main` function of `ament_flake8` return a subclass of `int` which also contains the actual error strings.

On the one hand it maintains backward compatibility with existing tests which compare the value with `== 0`. On the other hand it allows tests to also show the actual error messages, e.g. as part of the assert message which will be visible in Jenkins for a failed test (see referenced PR).

### DIFF
--- a/ament_flake8/ament_flake8/main.py
+++ b/ament_flake8/ament_flake8/main.py
@@ -126,7 +126,7 @@ def main(argv=sys.argv[1:]):
         with open(args.xunit_file, 'w') as f:
             f.write(xml)
 
-    return rc
+    return ReturnCode(rc, errors=report.errors)
 
 
 def get_flake8_style_guide(argv):
@@ -178,10 +178,7 @@ def generate_flake8_report(config_file, paths, excludes, max_line_length=None):
         format_func(error)
         report.add_error(error)
         print('')
-        print(
-            '%s:%d:%d: %s %s' % (
-                error.filename, error.line_number,
-                error.column_number, error.code, error.text))
+        print(format_error(error))
     style._application.formatter.format = custom_format
 
     # Get the names of files checked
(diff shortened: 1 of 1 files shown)


ament_flake8/ament_flake8/main.py
```diff
@@ -285,5 +288,21 @@ def print_statistics(self):
         self.report._application.report_statistics()
 
 
+class ReturnCode(int):
```
ivanpauno: Adding a `__repr__` method can maybe be useful.

dirk-thomas: What do you propose that method to return?

ivanpauno: I imagine this being used like:
```python3
if ret_code != 0:
    print(ret_code)
```

So, I would return something like:
```python3
def __repr__(self):
        return f"ReturnCode(): code='{str(super())}' error='{self.error_strings}'"
```

dirk-thomas:
Since packages using this API to test their code need to be updated to show the errors it is much cleaner to simply expose the errors through a new API without obfuscating the return code like in the first commit.

Therefore looking for another review.

dirk-thomas:
Moving forward without a review. Please feel free to comment after the fact and I will be happy to reply.

ament_flake8/ament_flake8/main.py
```diff
@@ -193,6 +195,12 @@ def custom_format(error):
     return report
 
 
+def format_error(error):
+    return '%s:%d:%d: %s %s' % (
```
ivanpauno: nit: Use f-strings. To be fair, this was moved from pre-existing code.

PR #226, opened 2020-03-24
some more clang format options

Karsten1987:
a few changes in how constructor initializer lists are formatted, 2 spaces before trailing comment, sorting `#includes` and auto completion for `namespace` ending.

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/ament_clang_format/ament_clang_format/configuration/.clang-format
+++ b/ament_clang_format/ament_clang_format/configuration/.clang-format
@@ -16,4 +16,15 @@ ContinuationIndentWidth: 2
 DerivePointerAlignment: false
 PointerAlignment: Middle
 ReflowComments: false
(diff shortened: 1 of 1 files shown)


ament_clang_format/ament_clang_format/configuration/.clang-format
```diff
@@ -16,4 +16,15 @@ ContinuationIndentWidth: 2
 DerivePointerAlignment: false
 PointerAlignment: Middle
 ReflowComments: false
+SpaceBeforeCtorInitializerColon: true
```
dirk-thomas: This seems to be the default value for the Google style already. So it shouldn't need to be set.

Same for several other options.

Karsten1987:
In fact, all of these values are already in the Google default style. Don't know what happened here :D

PR #229, opened 2020-04-14
Increase timeout for cppcheck runs.

nuclearsandwich:
On some platforms the cppcheck for the test_rclcpp package is hitting
the timeout. Bumping the timeout on those platforms shows that the build
takes about 130 seconds. So 180 should be sufficient room to grow
without raising it too high.

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
@@ -55,7 +55,7 @@ function(ament_cppcheck)
   ament_add_test(
     "${ARG_TESTNAME}"
     COMMAND ${cmd}
-    TIMEOUT 120
+    TIMEOUT 180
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
CI run with just Fast-RTPS. Build up to test_rclcpp and run just the cppcheck test for test_rclcpp

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10174/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5729/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8265/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10056/)

nuclearsandwich:
CI is yellow with known warnings on master. Going ahead and merging.

nuclearsandwich:
I actually messed this up. macOS is still timing out at 180s. It's even in the CI output above and I didn't see it past the scrolloff. First step is figuring out how long its taking on macOS and trying to figure out why.

PR #232, opened 2020-04-28
Allow AMENT_IGNORE markers to be directories

rotu:
This matches `COLCON_IGNORE`, which checks that a file *exists*, not necessarily that it is a normal file.


dirk-thomas:
Pending the question discussed on https://github.com/ros-infrastructure/catkin_pkg/pull/286#issuecomment-620880126

dirk-thomas:
With the REP update merged (ros-infrastructure/rep#256) I will go ahead ad merge this.

PR #234, opened 2020-04-30
Updated cpplint and cppcheck to exclude directories and files

ahcorde:
As we discussed in this PR https://github.com/ros2/geometry2/pull/258#issuecomment-621499030, it makes sense to exclude from cppcheck or cpplint some of the files or directories in a package.

We can use these arguments:
 - ament_cppcheck
```
  ament_cppcheck(LANGUAGE "c++"
    EXCLUDE_DIRS
      "include/tf2/LinearMath/* include/tf2/impl/*"
    EXCLUDE_FILES
       include/tf2/LinearMath/Scalar.h
       include/tf2/LinearMath/MinMax.h
  )
```
 - ament_lint
```
  ament_cpplint(EXCLUDE_DIRS include/tf2/LinearMath/)
```

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #238, opened 2020-05-11
Exclude files or directories in cpplint.py

ahcorde:
This PR is related to this other https://github.com/ament/ament_lint/pull/234

`ament_cmake_cpplint` includes all the files in the package, with this option we can exclude some files or directories.

Signed-off-by: ahcorde <ahcorde@gmail.com>


ament_cpplint/ament_cpplint/cpplint.py
```diff
@@ -141,6 +142,12 @@
       Examples:
         --linelength=120
 
+    exclude=dir1 dir2 file1...
+      List of exclude directories or files for cpplint
+
+      Examples:
+        --excludedir=LinearMath impl
+
```
dirk-thomas, line 150: This comment (https://github.com/ament/ament_lint/pull/234#issuecomment-625991937) still applies.

ahcorde: Done https://github.com/ament/ament_lint/pull/238/commits/b9f07a6de02a3eca53bd82b4562b7748dcb14166

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,7 +148,8 @@ def main(argv=sys.argv[1:]):
         else:
             print("Not using '--root'")
         print('')
-        arguments += files
+        arguments += [file for file in files
```
dirk-thomas, line 151: Nitpick: `file` is a builtin and shouldn't be used as a variable name (I would guess this also results in a linter warning).

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,7 +148,8 @@ def main(argv=sys.argv[1:]):
...
+                      if any(excl not in file or not excl for excl in args.exclude)]
```
dirk-thomas, line 152: Nitpick: according to the style guide to not align wrapped line but instead wrap directly after the open parenthesis / bracket and indent one level.

dirk-thomas, line 152: Can you explain the logic of this condition? `excl not in file` looks wrong to me.

ahcorde: `if excl is not in the file` -> this means that if the `excl` file or path is not in the the "include" filenames or paths then should **not** be discarded

ahcorde: you mean:

```python
      arguments += [file for file in files  if any(
          excl not in file or not excl for excl in args.exclude)]
```

dirk-thomas:
```
arguments += [
    f for f in files
    if any(excl not in f or not excl for excl in args.exclude)]
```

dirk-thomas: So `file` (which is btw a `builtin` and therefore shouldn't be used as a variable name) is the path of a to-be-checked file, right? E.g. `foo/bar/baz.ext`.

Why should an `excl` value `bar` skip that file? When passing `--exclude x` that would exclude every file which contains an `x`?

ahcorde: > Why should an excl value bar skip that file? 

I was trying to simplify it. Instead of writing the whole path, just writing the folder o the filename

> When passing --exclude x that would exclude every file which contains an x?
yes, you are right

ament_cpplint/ament_cpplint/main.py
```diff
@@ -77,6 +77,10 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--root', type=str,
         help='The --root option for cpplint')
+    parser.add_argument(
+        '--exclude', default=[''],
```
dirk-thomas, line 81: The default value should be an empty list and not contain an empty string. That also avoids the condition below `not excl`.

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,7 +148,9 @@ def main(argv=sys.argv[1:]):
         else:
             print("Not using '--root'")
         print('')
-        arguments += files
+        arguments += [
+            f for f in files
+            if any(excl not in f or not excl for excl in args.exclude)]
```
dirk-thomas, line 153: This logic is still wrong. You want to check `if f not in args.excludes`.

ahcorde: `args.exclude` constains substring of `f`. This condicion is not valid `if f not in args.excludes`.

`args.exclude` could contain directories which make the condition is not valid.

``
f = /tmp/ros2_foxy/src/ros2/geometry2/tf2/test/cache_unittest.cpp
args.excludes = test
```

In this case using this condition f is not in args.excludes. This is True when we want False (to not include this file)

but `any(excl not in f for excl in args.exclude)`  the substring is compared with the full name which will return True when the substring it not in the filename (adding this file to the cpplint argument list) and False when it is (removing it from the argument list)

File: ament_cpplint/ament_cpplint/main.py [hunk 6]
```diff
@@ -77,6 +77,10 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--root', type=str,
         help='The --root option for cpplint')
+    parser.add_argument(
+        '--exclude', default=[],
```
dirk-thomas, line 81: Nitpick: use plural `excludes` since multiple arguments are accepted.

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,7 +148,7 @@ def main(argv=sys.argv[1:]):
         else:
             print("Not using '--root'")
         print('')
-        arguments += files
+        arguments += [f for f in files if any(excl not in f for excl in args.exclude)]
```
dirk-thomas, line 151: As I have explained above with an example `in` / substring is not a sufficient check for what you want to do. In order to match exact files as well as directories as described in the help string you need to check for `f in args.exclude` or if any of the excludes matches a parent path of `f`.

ament_cpplint/ament_cpplint/main.py
```diff
@@ -148,7 +148,9 @@ def main(argv=sys.argv[1:]):
         else:
             print("Not using '--root'")
         print('')
-        arguments += [f for f in files if any(excl not in f for excl in args.exclude)]
+        arguments_dirs = [f for f in files if os.path.dirname(f) not in args.excludes]
```
dirk-thomas, line 151: This logic still doesn't cover the case where `--excludes` contain `foo` and a file `foo/bar/baz.h` should be skipped.

ahcorde: Improved logic. https://github.com/ament/ament_lint/pull/238/commits/63d09b5f12884877d114de0d1ef8120d8f745815 
I added a condition If you decided to exclude a complete folder cpplnt should not execute this block of code. because cpplint is being called  by folders. First include, then src, etc... Sumarizing the script will return a failured test if there is/are no file/files to check without this condition

ahcorde: friendly ping @dirk-thomas . Running CI for tf2 test avoiding `Linearmath` on cpplint, cppcheck and uncrustify https://github.com/ros2/geometry2/pull/258#issuecomment-636096343

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,27 +148,29 @@ def main(argv=sys.argv[1:]):
...
-                        'confidence': confidence,
-                        'message': message,
-                    })
-                DefaultError(filename, linenum, category, confidence, message)
-            ament_cpplint.cpplint.Error = custom_error
-
-            ProcessFile(filename, _cpplint_state.verbose_level)
-            report.append((filename, errors))
-            print('')
+        files_check = [f for f in files if not any(f for excl in args.excludes if excl in f)]
```
dirk-thomas, line 151: As commented multiple times before I don't see how `excl in f` is the appropriate condition here.

ahcorde: - The variables files contains all the files to check by cpplint. 
 - In args.excludes I have the excluded files or directories. 

Then I check if there is one of the excludes in the file with `any(f for excl in args.excludes if excl in f)` this will return the opposite condition that we want. That's why I negate this with the **not**. 

example: 

files = [`foo/file.h`, `foo/file2.h`, `bar/file.h`]
excludes = ['bar']

any(bar in `foo/file.h`) -> False -> not any() -> True -> we will check this file
any(bar in `foo/file2.h`) -> False -> not any() -> True -> we will check this file
any(bar in `bar/file.h`) -> True -> not any() -> False -> we will not check this file

dirk-thomas: I havfiles = ['foo/bar/baz']e given you examples multiple time why this *doesn't* work. Just try it with:

```
files = ['foo/bar/baz']
args.excludes = ['bar']
```

The result of `files_check` is empty which is incorrect.

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,27 +148,29 @@ def main(argv=sys.argv[1:]):
...
+        if(len(files_check) != 0):
```
ahcorde, line 152: @dirk-thomas  I agree with your comment https://github.com/ament/ament_lint/pull/238#discussion_r436100214, but I added this condition to avoid this case

dirk-thomas: How is this condition going to help the problem that the *wrong* files are being excluded in the line above?

ahcorde: your example is a valid case, it's when you want to exclude a hole folder.

files = ['foo/bar/baz']
args.excludes = ['bar']

files_check in this case is empty, The condition allows us to avoid running this test

dirk-thomas: No, in that example the directory `foo/bar/baz` should *not* be excluded since the excludes only list the name `bar` which should only match a file / directory which is a sibling of `foo`.

To make the example even more obvious:

```
files = ['foobarbaz']
args.excludes = ['bar']
```

Your logic does exclude the file / directory `foobarbaz` which is wrong.

ahcorde: I modified the logic https://github.com/ament/ament_lint/pull/238/commits/66ed86bd3baf3ebd942572ae0adefc4ea21b3f93 

I get the file_name and tokenize the path. Then if one of the excludes match with file_name or the path_token I add this file to an avoid file list. Finally I remove the common elements to get only the files to test.

ahcorde: I included pathlib to check with all the parent if the there are any matches https://github.com/ament/ament_lint/pull/238/commits/31ea18d59ed3ad0e359bdc64511c6a670d8ac24f

ament_cpplint/ament_cpplint/main.py
```diff
@@ -144,27 +149,49 @@ def main(argv=sys.argv[1:]):
...
+            for excl in args.excludes:
+                if (
+                    excl == file_name or
+                    path_to_check.match(excl) or
+                    check_path_parents(path_to_check.parents, excl)
+                ):
+                    files_to_avoid.append(f)
+        files_check = list(set(files) ^ set(files_to_avoid))
+
+        if(len(files_check) != 0):
```
dirk-thomas, line 173: This check us only necessary since the previous check from line 123 isn't effective since the exclude logic is applied afterwards. How about moving the logic into `get_file_groups()` - which I think would be the most efficient approach since it even avoid crawling / collecting files under an excluded directory. I would expect that would also avoid the need to check all `parents`.

Nitpick: no parenthesis around Python conditions. The Pythonic way to write this condition would be: `if files_check`.

ahcorde: Logic moved to `get_file_groups()` https://github.com/ament/ament_lint/pull/238/commits/3b9b453dc0e4361ff7c94a6f054fd91da8f6d444

dirk-thomas: The `exclude` logic can already be applied to the `dirnames` variable to filter them out early (without the need to recurse into them). That will also simplify the check later which doesn't have to consider parents.

ahcorde: moved to dirnames https://github.com/ament/ament_lint/pull/238/commits/3b9b453dc0e4361ff7c94a6f054fd91da8f6d444

dirk-thomas: I don't see where the code filters `dirnames` based on the `excludes`?

ament_cpplint/ament_cpplint/main.py
```diff
@@ -199,27 +205,59 @@ def custom_error(filename, linenum, category, confidence, message):
...
+
                 if 'AMENT_IGNORE' in dirnames + filenames:
                     dirnames[:] = []
                     continue
                 # ignore folder starting with . or _
                 dirnames[:] = [d for d in dirnames if d[0] not in ['.', '_']]
                 dirnames.sort()
+                for dirname in dirnames:
+                    if is_file_or_directory_to_exclude(os.path.join(dirpath, dirname), excludes):
+                        continue
```
dirk-thomas, line 249: This loop doesn't do anything. It needs to remove excluded directories from `dirnames` instead.

Also there shouldn't be a need for a recursive check like `is_file_or_directory_to_exclude`.

dirk-thomas: You just requested another review on this PR but I don't see any update after this comment?

ahcorde: I pushed in the wrong remote  https://github.com/ament/ament_lint/pull/238/commits/201486a59d695a9b304187d7c739584ee8640e99

ament_cpplint/ament_cpplint/main.py
```diff
@@ -199,27 +205,57 @@ def custom_error(filename, linenum, category, confidence, message):
...
             for dirpath, dirnames, filenames in os.walk(path):
+                if is_file_or_directory_to_exclude(dirpath, excludes):
+                    continue
+
                 if 'AMENT_IGNORE' in dirnames + filenames:
                     dirnames[:] = []
                     continue
                 # ignore folder starting with . or _
                 dirnames[:] = [d for d in dirnames if d[0] not in ['.', '_']]
+                dirnames[:] = [d for d in dirnames if d not in excludes]
```
dirk-thomas, line 246: `dirnames` only contains the names of the subdirectories of `dirpath`. As such the condition `not in excludes` is incorrect again - for the same reasons as the previous comments.

ament_cpplint/ament_cpplint/main.py
```diff
@@ -199,27 +205,60 @@ def custom_error(filename, linenum, category, confidence, message):
...
+                ]
 
+                dirnames.sort()
                 # select files by extension
                 for filename in sorted(filenames):
                     _, ext = os.path.splitext(filename)
                     if ext in ('.%s' % e for e in extensions):
-                        append_file_to_group(groups,
-                                             os.path.join(dirpath, filename))
+                        if filename not in excludes:
```
dirk-thomas, line 256: This condition still doesn't make sense.

Please write a unit test which tests a variety of cases to validate the implementation against.

ahcorde: humm, why not? the excludes argument might contains a filename

dirk-thomas: I don't think an exclude of `bar` should match a file `foo/bar`. Instead the exclude should state `foo/bar` explicitly. Otherwise how do I exclude the file `bar` but not the file `foo/bar`?

ahcorde: should I need to include the test?

dirk-thomas: You don't have to include the test in this PR but as mentioned before writing one will help you to check the logic. Especially the part to consider example file system structure and excludes and what the filtered result is expected to be.

ahcorde: @dirk-thomas  I created a gist with a test: https://gist.github.com/ahcorde/bcc558aedad8560cfa46c4b5de366f32

This is the last commit https://github.com/ament/ament_lint/pull/238/commits/6db3ae3adaf9cda7a11dfe23fe279f18c702fb7c

dirk-thomas: I am not sure how to use the gist. `get_file_groups()` seems to always return an empty dict?

Just from reading over the logic I would expect that a file `include/test/cache_unittest.cpp` would be incorrectly skipped even though only `test/cache_unittest.cpp` is specified in the excludes.

Also I still don't see why `is_file_or_directory_to_exclude()` should needs any logic like `check_path_parents()`.

JWhitleyWork:
@ahcorde Any update on this?

mm318:
I agree with Dirk on https://github.com/ament/ament_lint/pull/238#discussion_r458953121

> I don't think an exclude of `bar` should match a file `foo/bar`. Instead the exclude should state `foo/bar` explicitly. Otherwise how do I exclude the file `bar` but not the file `foo/bar`?

If the desire is to match `bar` at any level, then the expression should be something like `**/bar` (basically glob rules). Python already has a glob library as part of its standard libraries (https://docs.python.org/2.7/library/glob.html), we can just leverage that.

mm318:
The question now is what's the behavior already built into `ament_cppcheck` at
https://github.com/ament/ament_lint/blob/44da06bf85129ba5a2f159bff4fb00731126daf6/ament_cppcheck/ament_cppcheck/main.py#L150

audrow:
@ahcorde, what do you think we should do here? Should we close this PR?

ahcorde:
yes, we can close this PR

PR #241, opened 2020-05-12
Expand ignores to pep257 definition.

### LINKED ISSUE #240: ament_pep257 does not match pydocstyle's pep257 definition
The set of ignores that defines pep257 in pydocstyle http://www.pydocstyle.org/en/latest/error_codes.html#default-conventions
```
['D203', 'D212', 'D213', 'D214', 'D215', 'D404', 'D405', 'D406', 'D407', 'D408', 'D409', 'D410', 'D411', 'D413', 'D415', 'D416', 'D417']
```
versus the set of default ignores in ament_pep257
```python
['D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107', 'D203', 'D212', 'D404']
```

Results in the following checks that are not part of pep257 being run
```
['D213', 'D214', 'D215', 'D405', 'D406', 'D407', 'D408', 'D409', 'D410', 'D411', 'D413', 'D415', 'D416', 'D417']
```

pydocstyle generates the list of codes to check by either using a convention or an ignore list. ament_pep257 currently uses `--ignore`
https://github.com/ament/ament_lint/blob/db91e6cdfd719be2ed18b15f7e7cbbf0d609b4d6/ament_pep257/ament_pep257/main.py#L120

This should probably be changed to `--add-ignore`. Then, pydocstyle will default to using `--convention=pep257`, and then pep257's defaults are used in addition to any user specified ignores.

We can simplify the default ignores at that point to just `['D1']`
https://github.com/ament/ament_lint/blob/db91e6cdfd719be2ed18b15f7e7cbbf0d609b4d6/ament_pep257/ament_pep257/main.py#L48-L51

Arnatious:
Fixes #240 by using `--add-ignore` to allow user to expand pep257 ignores, and add error codes to check.

`-add-select` functionality added via the `--select` flag, allowing full customization of error codes to check

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -38,6 +38,10 @@ except ImportError:  # try version 1.0.0
 log.setLevel(logging.INFO)
 
 
+_conventions = set(pydocstyle.conventions.keys())
+_conventions.add('ament')
+
+
 def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
         description='Check docstrings against the style conventions in PEP 257.',
@@ -58,9 +62,9 @@ def main(argv=sys.argv[1:]):
     )
     err_code_group.add_argument(
         '--convention',
-        choices=pydocstyle.conventions,
+        choices=_conventions,
         default='ament',
-        help=f'Choose a preset list of error codes. Valid options are {pydocstyle.conventions}'
+        help=f'Choose a preset list of error codes. Valid options are {_conventions}'
     )
     parser.add_argument(
         '--add-ignore',
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -41,6 +41,20 @@ log.setLevel(logging.INFO)
 _conventions = set(pydocstyle.conventions.keys())
 _conventions.add('ament')
 
+_ament_ignore = [
+    'D100',
+    'D101',
+    'D102',
+    'D103',
+    'D104',
+    'D105',
+    'D106',
+    'D107',
+    'D203',
+    'D212',
+    'D404',
+]
+
 
 def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
@@ -64,7 +78,10 @@ def main(argv=sys.argv[1:]):
         '--convention',
         choices=_conventions,
         default='ament',
-        help=f'Choose a preset list of error codes. Valid options are {_conventions}'
+        help=(
+            f'Choose a preset list of error codes. Valid options are {_conventions}.'
+            f'The "ament" convention is defined as --ignore {_ament_ignore}.'
+        ),
     )
     parser.add_argument(
         '--add-ignore',
@@ -107,21 +124,7 @@ def main(argv=sys.argv[1:]):
     args.add_select = ','.join(args.add_select)
     args.add_ignore = ','.join(args.add_ignore)
     if not (args.ignore or args.select) and args.convention == 'ament':
-        args.ignore = ','.join(
-            [
-                'D100',
-                'D101',
-                'D102',
-                'D103',
-                'D104',
-                'D105',
-                'D106',
-                'D107',
-                'D203',
-                'D212',
-                'D404',
-            ]
-        )
+        args.ignore = ','.join(_ament_ignore)
 
     excludes = [os.path.abspath(e) for e in args.excludes]
     report = generate_pep257_report(args.paths, excludes, args.ignore, args.select,
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -42,18 +42,38 @@ def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
         description='Check docstrings against the style conventions in PEP 257.',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    parser.add_argument(
+    err_code_group = parser.add_mutually_exclusive_group()
+    err_code_group.add_argument(
         '--ignore',
         nargs='+',
         action='extend',
         default=[],
-        help='The pep257 error codes to ignore. Has precedence over --select.')
-    parser.add_argument(
+        help='Choose the list of error codes for pydocstyle NOT to check for.')
+    err_code_group.add_argument(
         '--select',
         nargs='+',
         action='extend',
         default=[],
-        help='The pep257 error codes to check.'
+        help='Choose the basic list of error codes for pydocstyle to check for.'
+    )
+    err_code_group.add_argument(
+        '--convention',
+        choices=pydocstyle.conventions,
+        default='ament',
+        help=f'Choose a preset list of error codes. Valid options are {pydocstyle.conventions}'
+    )
+    parser.add_argument(
+        '--add-ignore',
+        nargs='+',
+        action='extend',
+        default=[],
+        help='Ignore an extra error code, removing it from the list set by --(select/ignore)')
+    parser.add_argument(
+        '--add-select',
+        nargs='+',
+        action='extend',
+        default=[],
+        help='Check an extra error code, adding it to the list set by --(select/ignore).'
     )
     parser.add_argument(
         'paths',
@@ -75,19 +95,29 @@ def main(argv=sys.argv[1:]):
         help='Generate a xunit compliant XML file')
     args = parser.parse_args(argv)
 
-    default_ignore = set(['D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107'])
-    if 'D1' in args.select:
-        default_ignore = set([])
-    default_ignore -= set(args.select)
-    args.ignore += default_ignore
-
-    args.select.append('D213')
-
     if args.xunit_file:
         start_time = time.time()
 
+    args.ignore = ','.join(args.ignore)
+    args.select = ','.join(args.select)
+    args.add_select = ','.join(args.add_select)
+    args.add_ignore = ','.join(args.add_ignore)
+    if not (args.ignore or args.select) and args.convention == 'ament':
+        args.ignore = ','.join(['D100',
+                                'D101',
+                                'D102',
+                                'D103',
+                                'D104',
+                                'D105',
+                                'D106',
+                                'D107',
+                                'D203',
+                                'D212',
+                                'D404'])
+
     excludes = [os.path.abspath(e) for e in args.excludes]
-    report = generate_pep257_report(args.paths, excludes, args.ignore, args.select)
+    report = generate_pep257_report(args.paths, excludes, args.ignore, args.select,
+                                    args.convention, args.add_ignore, args.add_select)
     error_count = sum(len(r[1]) for r in report)
 
     # print summary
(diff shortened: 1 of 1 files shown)


Arnatious:
> 
> 
> I see the interest in changing `ament_pep257` to actually comply with PEP257, but what's the harm in the former performing a few more checks?

At least in the case of D212 and D213 - the two checks are mutually exclusive and opinionated (start multiline docstrings on next line vs start on same line as quotes).

The remainder are fine to keep, though some are redundant.

ament_pep257/ament_pep257/main.py
```diff
@@ -45,11 +45,14 @@ def main(argv=sys.argv[1:]):
...
-            'D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107',
-            'D203', 'D212', 'D404',
-        ],
+        default=['D1'],
         help='The pep257 categories to ignore')
+    parser.add_argument(
+        '--select',
+        nargs='*',
+        default=[],
+        help='The additional pydocstyle checks to run'
```
kyrofa, line 54: Given that `--ignore` turns into using `--add-ignore` and `--select` turns into using `--add-select`, I think the terminology for the help text should be the same. `--ignore` is using the term "pep257 categories", `--select` is using "pydocstyle checks", and the [pydocstyle docs](http://www.pydocstyle.org/en/5.0.2/usage.html#command-line-interface) uses the term "error codes to check." Can we pick one and go with it, with a personal preference to upstream's term?

ament_pep257/ament_pep257/main.py
```diff
@@ -45,11 +45,14 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--ignore',
         nargs='*',
-        default=[
-            'D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107',
-            'D203', 'D212', 'D404',
-        ],
+        default=['D1'],
```
kyrofa, line 48: Given the point of this PR, it's worth mentioning that leaving this as the default still means this isn't exactly pep257.

Arnatious: 0b9856b added a flag and a warning about it being required in the future so as to not break current tests

ament_pep257/ament_pep257/main.py
```diff
@@ -75,6 +82,13 @@ def main(argv=sys.argv[1:]):
 
     if args.xunit_file:
         start_time = time.time()
+    if args.allow_undocumented:
+        args.ignore.append('D1')
+    elif args.allow_undocumented is None:
+        warnings.warn('Argument "--allow-undocumented" will be required in a future version. '
+                      'Defaulting to --allow-undocumented=True')
+        args.ignore.append('D1')
```
kyrofa, line 90: This is excellent, nice work.

ament_pep257/ament_pep257/main.py
```diff
@@ -45,11 +46,20 @@ def main(argv=sys.argv[1:]):
...
+        default=[],
+        help='The pep257 error codes to ignore.')
+    parser.add_argument(
+        '--select',
+        nargs='*',
+        default=[],
+        help='The pep257 error codes to check.'
+    )
+    parser.add_argument(
+        '--allow-undocumented',
```
kyrofa, line 58: I kinda feel like this should be `action='store_true'`, but that might interfere with the backward compatibility check. What do you think?

Arnatious: Yeah, there's no way to differentiate leaving out the argument from leaving it false in that case.

ament_pep257/ament_pep257/main.py
Arnatious: fixed in 0b9856b

ament_pep257/ament_pep257/main.py
Arnatious: Since the current behavior is contrary to pep257, we need a transition before adding this potentially breaking change.

ament_pep257/ament_pep257/main.py
```diff
@@ -45,11 +46,20 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--ignore',
         nargs='*',
-        default=[
-            'D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107',
-            'D203', 'D212', 'D404',
```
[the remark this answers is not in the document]
hidmic: Ultimately, an external user could relax constraints (i.e. ignore both D212 and D213) without us changing the defaults.

hidmic:
@Arnatious friendly ping.

ament_pep257/ament_pep257/main.py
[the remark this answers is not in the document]
Arnatious: Restored the default error checking behavior in e09982a.

Note that some merging code is necessary to duplicate the old behavior 

https://github.com/ament/ament_lint/blob/e09982a0e36865efe92dc4f6b53efe74dce8153f/ament_pep257/ament_pep257/main.py#L76-L82

sloretz:
@Arnatious @hidmic  friendly ping :) It looks like @Arnatious has made changes since your review, and now there's a merge conflict with this PR.

@Arnatious mind fixing the merge conflict?

Arnatious:
@sloretz rebased onto latest master. @hidmic ready for review

ament_pep257/ament_pep257/main.py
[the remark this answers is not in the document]
hidmic: Hmm, instead of making assumptions on what the default ignored error codes are and forcing others to set a baseline, I'd much rather replicate `pydocstyle` interface i.e. to have `--ignore`, `--select`, `--add-ignore`, and `--add-select`, and keep the default `--ignore` list as it used to be. Thoughts?

ament_pep257/ament_pep257/main.py
```diff
@@ -42,19 +42,44 @@ def main(argv=sys.argv[1:]):
...
+        nargs='+',
+        action='extend',
+        default=[],
+        help='Choose the basic list of error codes for pydocstyle to check for.'
+    )
+    err_code_group.add_argument(
+        '--convention',
+        choices=pydocstyle.conventions,
+        default='ament',
+        help=f'Choose a preset list of error codes. Valid options are {pydocstyle.conventions}'
```
hidmic, line 63: @Arnatious will `ament` show up in `pydocstyle.conventions`? I'd expect `ament_pep257 --convention=ament` not to fail, even if redundant.

ament_pep257/ament_pep257/main.py
```diff
@@ -73,8 +98,26 @@ def main(argv=sys.argv[1:]):
...
+                                'D101',
+                                'D102',
+                                'D103',
+                                'D104',
+                                'D105',
+                                'D106',
+                                'D107',
+                                'D203',
+                                'D212',
+                                'D404'])
```
hidmic, line 116: @Arnatious uh, is `flake8` happy with this?

ament_pep257/ament_pep257/main.py
Arnatious: a4d6399 added logic to populate/allow "ament" as an option

ament_pep257/ament_pep257/main.py
Arnatious: It was, but I ran it through black in 6a645fc and it looks better now

hidmic:
@Arnatious mind to fix DCO? I'll run CI afterwards.

Arnatious:
@hidmic fixed

dirk-thomas:
How can a user get the information what codes are being ignored (e.g. for the `ament` convention) without reading the source code?

hidmic:
Running CI up to `rclpy`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11381/) (canary build)

Arnatious:
> 
> 
> How can a user get the information what codes are being ignored (e.g. for the `ament` convention) without reading the source code?

@dirk-thomas that's a good question. --convention is taken right from `pydocstyle`, where the only documentation of those error codes is at http://www.pydocstyle.org/en/latest/error_codes.html#default-conventions . 

ament is pep257, with the addition of ignoring "D1" (undocumented) and for some unexplained reason enforcing D213.

Would a readme explaining the options coming from pydocstyle/linking to the pydocstyle style docs as appropriate be desired?

dirk-thomas:
> Would a readme explaining the options coming from pydocstyle/linking to the pydocstyle style docs as appropriate be desired?

Before this patch a user could call `ament_pep257 --help` to get the information which I think is the most intuitive way (at least for the `ament` convention).

Arnatious:
@dirk-thomas added help text in e96b4ab 

Resulting output is 

```
usage: ament_pep257 [-h] [--ignore IGNORE [IGNORE ...] | --select SELECT
                    [SELECT ...] | --convention {google,pep257,numpy,ament}]
                    [--add-ignore ADD_IGNORE [ADD_IGNORE ...]]
                    [--add-select ADD_SELECT [ADD_SELECT ...]]
                    [--exclude [filename [filename ...]]]
                    [--xunit-file XUNIT_FILE]
                    [paths [paths ...]]

Check docstrings against the style conventions in PEP 257.

positional arguments:
  paths                 The files or directories to check. For directories,
                        files ending in '.py' will be considered. (default:
                        ['.'])

optional arguments:
  -h, --help            show this help message and exit
  --ignore IGNORE [IGNORE ...]
                        Choose the list of error codes for pydocstyle NOT to
                        check for. (default: [])
  --select SELECT [SELECT ...]
                        Choose the basic list of error codes for pydocstyle to
                        check for. (default: [])
  --convention {google,pep257,numpy,ament}
                        Choose a preset list of error codes. Valid options are
                        {'google', 'pep257', 'numpy', 'ament'}. The "ament"
                        convention is defined as --ignore ['D100', 'D101',
                        'D102', 'D103', 'D104', 'D105', 'D106', 'D107',
                        'D203', 'D212', 'D404']. (default: ament)
  --add-ignore ADD_IGNORE [ADD_IGNORE ...]
                        Ignore an extra error code, removing it from the list
                        set by --(select/ignore) (default: [])
  --add-select ADD_SELECT [ADD_SELECT ...]
                        Check an extra error code, adding it to the list set
                        by --(select/ignore). (default: [])
  --exclude [filename [filename ...]]
                        The filenames to exclude. (default: [])
  --xunit-file XUNIT_FILE
                        Generate a xunit compliant XML file (default: None)
```

dirk-thomas:
> added help text in e96b4ab

Thanks, LGTM.

hidmic:
Alright, CI's green, PR's approved. Going in. Thanks for your patience @Arnatious !

emersonknapp:
This PR seems to be causing runtime errors in our builds - e.g. https://github.com/ros-tooling/action-ros-ci/runs/870259711?check_suite_focus=true

```
  [3.105s] test/test_pep257.py F                                                    [100%]
  [3.105s] 
  [3.105s] =================================== FAILURES ===================================
  [3.105s] _________________________________ test_pep257 __________________________________
  [3.105s] test/test_pep257.py:22: in test_pep257
  [3.106s]     rc = main(argv=[])
  [3.106s] ../../../../build/ament_pep257/ament_pep257/main.py:69: in main
  [3.106s]     help='Choose the list of error codes for pydocstyle NOT to check for.')
  [3.106s] /usr/lib/python3.6/argparse.py:1346: in add_argument
  [3.106s]     raise ValueError('unknown action "%s"' % (action_class,))
  [3.106s] E   ValueError: unknown action "extend"
```

I'm not sure exactly what this issue is, but the `extend` action came from this PR

emersonknapp:
Followup - the `extend` action is new in Python 3.8, but ROS 2 Foxy supports Python 3.7 as per REP-2000 https://www.ros.org/reps/rep-2000.html#foxy-fitzroy-may-2020-may-2023

My recommendation is to revert this PR and then make sure we have a Python3.7-compliant build before merging it again.

kyrofa:
Indeed, extend is new of as 3.8.

Arnatious:
The behavior of extend isn't actually necessary, so I removed it in 7623ada, my bad @emersonknapp

emersonknapp:
Thanks for the followup - is that commit in a followup PR?

Arnatious:
> 
> 
> Thanks for the fix - is that commit in a followup PR?

Yes, https://github.com/ament/ament_lint/pull/262

PR #243, opened 2020-05-12
Remove match args to allow pydocstyle defaults

Arnatious:
Changes `--match` to use pydocstyle default

- Current: `.*\.py`
- New (Default): `(?!test_).*\.py`
Result: don't check 'test_*' files for documentation

Changes `--match-dir` to use pydocstyle default

- Current: `[^\._].*`
- New (Default): `[^\.].*`
Result: include private ('_*') package directories in checks

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -118,8 +118,6 @@ def generate_pep257_report(paths, excludes, ignore):
     sys.argv = [
         'main',
         '--ignore=' + ','.join(ignore),
-        '--match', r'.*\.py',
-        '--match-dir', r'[^\._].*',
     ]
     sys.argv += paths
     conf.parse()


dirk-thomas:
I rebased your branch since the target branch has changed quite a bit since the fork A first CI build on Linux to see how the modified config affect the result for the existing code: [Build Status](https://ci.ros2.org/job/ci_linux/11119/)

dirk-thomas:
> I also don't like the idea of code suddenly starting to fail the linter without the code changing.

This is certainly not something I would propose to be backported. But changing the default in the rolling distro is imo acceptable. Especially since the chances that it affects existing code is rather low (considering that there is not a single new warning due to this change in the whole ros2.repos file).

So I am planning to merge this within a day if there are no strong opinions not to accept the patch as is.

kyrofa:
> So I am planning to merge this within a day if there are no strong opinions not to accept the patch as is.

Not from me, you convinced me with:

> there is not a single new warning due to this change in the whole ros2.repos file

PR #247, opened 2020-05-15
Remove check of source files done in lint hook

Blast545:
As the title says, this will close #237. The comment was reworded as well to "source files" as this distinction is made on the `ament_copyright` linter.

 
Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

PR #248, opened 2020-05-18
Remove output on stderr for case that is not an error

Blast545:
This PR solves copyright test fails caused by #247. For the case of not a main repo file, the `ament_copyright` linter returned "no error" but printed info on `stderr` and didn't generate a result xml file. 

Failing nightly CI before the fix:
* https://ci.ros2.org/view/nightly/job/nightly_linux_release/1546

CI with packages that were reporting copyright errors before this fix:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10803/)

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>


dirk-thomas:
Regression from #247.

PR #242, opened 2020-05-12
support flake8 3.8 while maintaining compatibility with older versions

dirk-thomas:
CI builds only testing `ament_flake8` and `ament_index_python`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10707/)
  * Same parameters but testing with `master`: [Build Status](https://ci.ros2.org/job/ci_linux/10708/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6120/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8721/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10607/)

PR #252, opened 2020-06-01
[backport dashing] support flake8 3.8

emersonknapp:
Backport of #242 

Signed-off-by: Dirk Thomas <dirk-thomas@users.noreply.github.com>

### DIFF
--- a/ament_flake8/ament_flake8/main.py
+++ b/ament_flake8/ament_flake8/main.py
@@ -128,13 +128,26 @@ def get_flake8_style_guide(argv):
     # appropriate options to pass into the standard flake8.legacy.get_style_guide();
     # passing argv gets it to determine the options for us.
     application = flake8_app.Application()
-    application.parse_preliminary_options_and_args([])
-    flake8.configure_logging(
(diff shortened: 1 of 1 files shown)


emersonknapp:
@mjcarroll perhaps? Users who build Dashing from source via https://index.ros.org/doc/ros2/Installation/Dashing/Linux-Development-Setup/ will have all flake8 tests be broken without it

dirk-thomas:
Backport of #242. (@emersonknapp references in the ticker title don't cross reference.)

nuclearsandwich:
> LGTM, but I'm going to defer to @nuclearsandwich on this one (for Dashing).

Thanks for keeping me in the loop. For the future. I don't need to gatekeep Dashing backport PRs unless there's a deadlock or some sort of judgement call about breakages that needs to be made. It is important to make sure that backport PRs make it onto the patch release board (which this one is, even thought it's hard to tell because it's outside the ROS 2 org) so we make sure that backports also get releases but I don't feel the need to approve each and every one.

PR #253, opened 2020-06-01
[backport eloquent] support flake8 3.8

emersonknapp:
Signed-off-by: Dirk Thomas <dirk-thomas@users.noreply.github.com>


emersonknapp:
@mjcarroll perhaps? Users who build Eloquent from source via https://index.ros.org/doc/ros2/Installation/Dashing/Linux-Development-Setup/ will have all flake8 tests be broken without it.

Problem is that the `pip` install gets them flake8 3.8

dirk-thomas:
Backport of #242. (@emersonknapp references in the ticker title don't cross reference.)

PR #257, opened 2020-06-18
[Dashing] Fix bug, allows using ament_copyright with bsd3 (#206)

mjcarroll:
Backport of #206 for Dashing

Fix bug, allows using ament_copyright with bsd3

In the bs2 headers, exists a "{copyright_holder}" text that causes a problem
when using the command ament_copyright to add headers to a source file.

This fix adds a default value for that key, to match the original 3-Clause BSD
text, and allowing to use the tool.

Signed-off-by: Jorge J. Perez <jjperez@ekumenlabs.com>
Signed-off-by: Michael Carroll <michael@openrobotics.org>


sloretz:
@mjcarroll friendly ping; can this be merged now?

PR #260, opened 2020-06-22
Remove use of pkg_resources from ament_lint.

clalancette:
Replace it with the use of the more modern importlib_metadata
library.  There are a couple of reasons to do this:

1.  pkg_resources is quite slow to import; on my machine,
just firing up the python interpreter takes ~35ms, while
firing up the python interpreter and importing pkg_resources
takes ~175ms.  Firing up the python interpreter and importing
importlib_metadata takes ~70ms.  Removing 100ms per invocation
of the command-line both makes it speedier for users, and
will speed up our tests (which call out to the command-line
quite a lot).

2.  pkg_resources is somewhat deprecated and being replaced
by importlib.  https://importlib-metadata.readthedocs.io/en/latest/using.html
describes some of it

Note: By itself, this change is not enough to completely remove our
dependence on pkg_resources.  We'll also have to do something about
the console_scripts that setup.py generates.  That will be
a separate effort.

See ros2/ros2#919 for more details on why we want to do this.

### DIFF
--- a/ament_copyright/ament_copyright/__init__.py
+++ b/ament_copyright/ament_copyright/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pkg_resources
+import importlib_metadata
 
 
 COPYRIGHT_GROUP = 'ament_copyright.copyright_name'
@@ -35,8 +35,7 @@ UNKNOWN_IDENTIFIER = '<unknown>'
 
 def get_copyright_names():
     names = {}
-    for entry_point in pkg_resources.iter_entry_points(
-            group=COPYRIGHT_GROUP):
+    for entry_point in importlib_metadata.entry_points().get(COPYRIGHT_GROUP, []):
         assert entry_point.name != UNKNOWN_IDENTIFIER, \
             "Invalid entry point name '%s'" % entry_point.name
         name = entry_point.load()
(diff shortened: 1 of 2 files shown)


jacobperron:
Should this be considered for backport?

clalancette:
> Should this be considered for backport?

Good question.  There's no particular reason it cannot be.  However, the set of patches to remove pkg_resources usage in our code isn't enough to actually speed up command-line tools.  To actually get the speed-up, we'll also need to do some work to prevent pkg_resources from being used when we build the eggs for our command-line tools.  I think that is going to be more complicated, and may or may not be able to be backported.

So that's to say that I think we should hold off on thinking about backporting this until we know what the full solution looks like.  If you still think it would be worthwhile for backport, I can add it to the Foxy patch release board; it will probably just sit there while we figure out what we should do for the rest.  Just let me know.

jacobperron:
> If you still think it would be worthwhile for backport, I can add it to the Foxy patch release board

Considering we won't see an improvement, I think it's fine to sit on it until the related work is completed.

clalancette:
> Considering we won't see an improvement, I think it's fine to sit on it until the related work is completed. We don't need to add it to the Foxy board.

Sounds good, thanks.  Going ahead and merging this one now.

ament_copyright/ament_copyright/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pkg_resources
+import importlib_metadata
```
hidmic, line 15: @clalancette `importlib_metadata` is `importlib.metadata` on Python3.8 (see [here](https://importlib-metadata.readthedocs.io/en/latest/)), and so this script is failing for me on Windows.

clalancette: I added importlib-metadata in the `pip install` list for Rolling: https://index.ros.org/doc/ros2/Installation/Rolling/Windows-Development-Setup

It's somewhat redundant (since you can get it via `importlib.metadata` as you say), but it allows us to support Ubuntu 18.04 without a `try/except` block in the code.

PR #263, opened 2020-08-14
Increase cppcheck test timeout.

nuclearsandwich:
test_rclcpp is still seeing cppcheck timeouts on Windows.
This adds another 30 seconds to the timeout in order to provide more headroom.

Example builds that failed this week:

* https://ci.ros2.org/view/nightly/job/nightly_win_deb/1712/testReport/
* https://ci.ros2.org/view/nightly/job/nightly_win_rel/1660/testReport/

Builds running the cppcheck tests with retest-until-fail to increase the chance of getting a timeout:

* Windows debug [Build Status](https://ci.ros2.org/job/ci_windows/11843/) 
* Windows release [Build Status](https://ci.ros2.org/job/ci_windows/11883/)

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
@@ -61,7 +61,7 @@ function(ament_cppcheck)
   ament_add_test(
     "${ARG_TESTNAME}"
     COMMAND ${cmd}
-    TIMEOUT 180
+    TIMEOUT 210
     OUTPUT_FILE "${CMAKE_BINARY_DIR}/ament_cppcheck/${ARG_TESTNAME}.txt"
     RESULT_FILE "${result_file}"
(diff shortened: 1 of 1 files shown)


dirk-thomas:
For how long do these tests run when they don't reach the timeout? And what is the variability in their runtime? E.g. the first successful build would have passed without this change since it was 2 and a half minutes.

chapulina:
> For how long do these tests run when they don't reach the timeout? And what is the variability in their runtime?

It looks like it's usually less than a couple seconds for both `win_rel` and `win_deb`:

* https://ci.ros2.org/view/nightly/job/nightly_win_deb/1712/testReport/(root)/projectroot/cppcheck/history/
* https://ci.ros2.org/view/nightly/job/nightly_win_rel/1660/testReport/(root)/projectroot/cppcheck/history/

This suggests that increasing the timeout may not solve the issue. The test is probably hanging indefinitely.

PR #265, opened 2020-08-30
Simplify ament_lint include logic

rotu:
Previously the logic did not handle generator expressions in a consistent way and required users to add recursive dependencies to cppcheck e.g. https://github.com/ros2/rosidl_typesupport_connext/pull/57 . This should fix that. Performance impact should be assessed.


rotu:
Checking if this fixes https://github.com/ros2/rmw_cyclonedds/issues/220 
ROS 2 CI
https://github.com/rotu/ament_lint/actions/runs/230841540

PR #272, opened 2020-09-22
Allow correct languages on uncrustify

MiguelCompany:
When running `uncrustify -h` it does not show `C++` as a valid option for `-l`.

This PR changes the valid options on `--language` to `C` and `CPP`

Signed-off-by: Miguel Company <MiguelCompany@eprosima.com>

### DIFF
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -65,7 +65,7 @@ def main(argv=sys.argv[1:]):
         help='Exclude specific file names and directory names from the check')
     parser.add_argument(
         '--language',
-        choices=['C', 'C++'],
+        choices=['C', 'CPP'],
         help="Passed to uncrustify as '-l <language>' to force a specific "
              'language rather then choosing one based on file extension')
     parser.add_argument(
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -105,7 +105,7 @@ def main(argv=sys.argv[1:]):
             start_time = time.time()
 
         files_by_language = get_files(
-            args.paths, {'C': c_extensions, 'C++': cpp_extensions},
+            args.paths, {'C': c_extensions, 'CPP': cpp_extensions},
             excludes=args.exclude, language=args.language)
         if not files_by_language:
             print('No files found', file=sys.stderr)


claireyywang (approved):

mm318:
Is `ament_uncrustify` not translating `C++` to `CPP` when passing it to `uncrustify`?

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -65,7 +65,7 @@ def main(argv=sys.argv[1:]):
         help='Exclude specific file names and directory names from the check')
     parser.add_argument(
         '--language',
-        choices=['C', 'C++'],
+        choices=['C', 'CPP'],
```
mm318, line 68: I don't think the `C++` here is meant for `uncrustify`. I think it's meant for `ament_uncrustify`, and making this change will break the dictionary [here](https://github.com/ament/ament_lint/pull/272/files#diff-6e94bb1449472e0ebabe73155ff68bf9546cda13785946c5891a8afb741b9e04R108).

Can you double check?

mm318 (changes requested):

ament_uncrustify/ament_uncrustify/main.py
audrow: @mm318, can you confirm the link to the dictionary?

audrow: Nevermind, I see what you meant. 

I think you're right that it breaks the dictionary here.
https://github.com/ament/ament_lint/blob/d17140b716eb11f7cc8b3d686104ed2167ce9818/ament_uncrustify/ament_uncrustify/main.py#L107-L109

@MiguelCompany, the dictionary could be updated to use `CPP` and thus make `ament_uncrustify` more consistent with `uncrustify`. I believe line 108 is the only one that would need to change.

audrow:
After spending some time going through the code and trying `uncrustify`, I don't believe that this change breaks any dictionary.

@MiguelCompany I also think that the following line should also be changed:
https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L108

<details><summary>Following the code</summary>
<br>

The language is only used here:

https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L107-L109

`get_files` creates a dictionary where the language is the value. It optionally uses the supplied language.
https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L213-L217

This is ultimately used in grouping the files by language (file type is key, a list of files is the value):
https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L236-L239

The `uncrustify` is invoked on each group of languages:
https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L128-L129

The actual command used to call `uncrustify` then appends all of the files as the last argument:
https://github.com/ament/ament_lint/blob/2d0d0d1b29a5df003ad1f983a5a1fdc7d55f86e8/ament_uncrustify/ament_uncrustify/main.py#L269-L276
</details>

ament_uncrustify/ament_uncrustify/main.py
audrow: Nevermind.

audrow (commented):
Thanks for the PR @MiguelCompany. I think this is a good fix. 

I think that the change should also be added in one other place in the file. See https://github.com/ament/ament_lint/pull/272#issuecomment-786960345.

mm318:
So following the code, the value of `args.language` will be passed to `uncrustify`?

audrow:
> the value of args.language will be passed to uncrustify

Correct, as the language (`-l`), if it is supplied. If not, the file extensions will be used. If I understand correctly, `C++` isn't a valid language specifier for uncrustify (`CPP` is), but it still uses the supplied config file, so it seems to have worked.

MiguelCompany:
> Thanks for the PR @MiguelCompany. I think this is a good fix.
> 
> I think that the change should also be added in one other place in the file. See [#272 (comment)](https://github.com/ament/ament_lint/pull/272#issuecomment-786960345).

I rebased, fixed DCO and also updated the dictionary key from `C++` to `CPP`

audrow (approved):

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13770/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8636/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13848/)

audrow:
@mm318, does it look good to you, too?

mm318 (approved):
Yep, looks good to me!

audrow:
@MiguelCompany, thanks for the PR!

PR #271, opened 2020-09-08
Increase the ament_cppcheck timeout to 5 minutes.

clalancette:
This will avoid timeouts on some slower platforms that we've
started to see.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This fixes https://github.com/ros2/rclcpp/issues/1289

PR #277, opened 2020-11-06
[foxy backport] Increase the ament_cppcheck timeout to 5 minutes. (#271)

brawner:
This will avoid timeouts on some slower platforms that we've
started to see.

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cppcheck.cmake
@@ -61,7 +61,7 @@ function(ament_cppcheck)
   ament_add_test(
     "${ARG_TESTNAME}"
     COMMAND ${cmd}
-    TIMEOUT 180
+    TIMEOUT 300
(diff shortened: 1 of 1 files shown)


brawner:
Example of failure:
* MacOS [Build Status](https://ci.ros2.org/job/ci_osx/10710/)
With this PR:
* MacOS [Build Status](https://ci.ros2.org/job/ci_osx/10711/)

jacobperron (approved):

jacobperron:
@brawner the referenced example failure doesn't appear to be related.

brawner:
Oh weird, I ran into the error 3x in a row. I was certain it would happen again. This is an example of the error

[Build Status](https://ci.ros2.org/job/ci_osx/10709/)

brawner:
The timeout issue was resolved, but now there is a warning due to a recent commit to Fast-dds. I opened a PR for that issue at https://github.com/eProsima/Fast-DDS/pull/1558

brawner:
It looks like cppcheck took 3 minutes and 40 seconds to run, so 5 minute timeout looks about right.

PR #278, opened 2020-12-11
Use correct lint package dependencies

esteve:
This PR updates the documentation to match:

https://github.com/ament/ament_lint/blob/master/ament_cmake_cppcheck/doc/index.rst#how-to-run-the-check-from-within-a-cmake-ament-package-as-part-of-the-tests
https://github.com/ament/ament_lint/blob/master/ament_cmake_pycodestyle/doc/index.rst#how-to-run-the-check-from-within-a-cmake-ament-package-as-part-of-the-tests
https://github.com/ament/ament_lint/blob/master/ament_cmake_clang_format/doc/index.rst#how-to-run-the-check-from-within-a-cmake-ament-package-as-part-of-the-tests

according to these links, the packages that need to be added to `package.xml` seem to be `ament_cmake_${LINTER}`.

### DIFF
--- a/ament_lint_auto/doc/index.rst
+++ b/ament_lint_auto/doc/index.rst
@@ -24,9 +24,9 @@ test dependencies.
     <test_depend>ament_lint_auto</test_depend>
 
     <!-- add test dependencies on any linter, e.g. -->
-    <test_depend>ament_clang_format</test_depend>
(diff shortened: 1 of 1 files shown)


mm318:
Thank you for your contribution!

Issue #279, opened 2020-12-18
pycodestyle failure in flake8


tfoote:
The linux packaging job is failing due to what appears to be a flake8 API change

https://ci.ros2.org/view/packaging/job/packaging_linux/lastCompletedBuild/consoleText

`AttributeError: module 'pycodestyle' has no attribute 'break_around_binary_operator'`

<details>

```
4: Test command: /home/jenkins-agent/workspace/packaging_linux/venv/bin/python3 "-u" "/home/jenkins-agent/workspace/packaging_linux/ws/install/share/ament_cmake_test/cmake/run_test.py" "/home/jenkins-agent/workspace/packaging_linux/ws/build/ros1_bridge/test_results/ros1_bridge/flake8.xunit.xml" "--package-name" "ros1_bridge" "--output-file" "/home/jenkins-agent/workspace/packaging_linux/ws/build/ros1_bridge/ament_flake8/flake8.txt" "--command" "/home/jenkins-agent/workspace/packaging_linux/ws/install/bin/ament_flake8" "--xunit-file" "/home/jenkins-agent/workspace/packaging_linux/ws/build/ros1_bridge/test_results/ros1_bridge/flake8.xunit.xml"
4: Test timeout computed to be: 60
4: -- run_test.py: invoking following command in '/home/jenkins-agent/workspace/packaging_linux/ws/src/ros2/ros1_bridge':
4:  - /home/jenkins-agent/workspace/packaging_linux/ws/install/bin/ament_flake8 --xunit-file /home/jenkins-agent/workspace/packaging_linux/ws/build/ros1_bridge/test_results/ros1_bridge/flake8.xunit.xml
4: Traceback (most recent call last):
4:   File "/home/jenkins-agent/workspace/packaging_linux/venv/lib/python3.6/site-packages/flake8/plugins/manager.py", line 157, in load_plugin
4:     self._load()
4:   File "/home/jenkins-agent/workspace/packaging_linux/venv/lib/python3.6/site-packages/flake8/plugins/manager.py", line 134, in _load
4:     self._plugin = self.entry_point.load()
4:   File "/home/jenkins-agent/workspace/packaging_linux/venv/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 98, in load
4:     return functools.reduce(getattr, attrs, module)
4: AttributeError: module 'pycodestyle' has no attribute 'break_around_binary_operator'
4: 
4: During handling of the above exception, another exception occurred:
4: 
[...]

PR #281, opened 2021-01-04
Remove dependency on broken package pyhon3-flake8

tylerjw:

This should fix this issue: https://github.com/ament/ament_lint/issues/279

The problem is the released versions of python3-flake8 in debians is broken for now because it doesn't handle the fact that the latest verion of flake8 depends on a newer version of pycodestyle.  If you instead isntall flake8 using pip3 like this: `pip3 install flake8` it works just fine.  I don't know how pip dependencies are defined for CI so this may not be a complete solution.  It did enable me to use this from source locally though.

### DIFF
--- a/ament_flake8/package.xml
+++ b/ament_flake8/package.xml
@@ -11,7 +11,6 @@
   <author email="dhood@osrfoundation.org">D. Hood</author>
 
   <exec_depend>ament_lint</exec_depend>
-  <exec_depend>python3-flake8</exec_depend>
 
   <export>
     <build_type>ament_python</build_type>


tylerjw:
As I suspected this didn't pass tests as the dependency of flake8 is now missing.  Maybe we should fix this on the rosdep side and have it depend on the pip version of flake8.  Is that possible, or can you only depend on debians?

mm318:
> As I suspected this didn't pass tests as the dependency of flake8 is now missing. Maybe we should fix this on the rosdep side and have it depend on the pip version of flake8. Is that possible, or can you only depend on debians?

Yeah, it would be better to fix it on the `rosdep` side. The fix should be made somewhere [here](https://github.com/ros/rosdistro/blob/f5a1e860b602a2e91356386f9b492460dc0c482a/rosdep/python.yaml#L5810). Neighboring entries show how to depend on `pip` instead of `apt`.

audrow:
@tylerjw, any updates on this? Do you want to open a PR on rosdistro? FYI, I believe [this](https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#python-3) is the relevant section of the rosdistro contribution guide.

audrow:
I'm going to close this PR as #279 has been closed.

PR #282, opened 2021-01-04
Add --clang-format-version option

tylerjw:

## Description
I am using clang-format-10.  This change would enable me to use ament_clang_format with clang-format-10.  Note that `find_executable` will return as soon as it finds a valid executable.  This means that if we search in this order searching for `clang-format-` first causes no harm in the case where the user does not provide a value for this.

### Follow-on work (for other PRs)
* `ament_cmake_clang_format` should be updated to enable using this option from cmake.
* Would it be a good practice to search the filesystem for a .clang_format file in the working directory or in a parent and use that instead if it exists for the config?  This would simplify running ament_clang_format in a repo that contains a config.  If this default behavior would be accepted, it should also be done for clang-tidy.

### DIFF
--- a/ament_clang_format/ament_clang_format/main.py
+++ b/ament_clang_format/ament_clang_format/main.py
@@ -48,6 +48,9 @@ def main(argv=sys.argv[1:]):
              'in %s will be considered.' %
              ', '.join(["'.%s'" % e for e in extensions]))
     parser.add_argument(
+        '--clang-format-version',
+        help='The version of clang-format to use.')
+    parser.add_argument(
         '--reformat',
         action='store_true',
         help='Reformat the files in place')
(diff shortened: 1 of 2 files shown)


mm318:
Can you also please update the [doc](https://github.com/ament/ament_lint/blob/master/ament_clang_format/doc/index.rst) regarding the new option that you're adding?

tylerjw:
I updated the documentation to include instructions for setting these variables.  I will create a follow-on PR for `ament_cmake_clang_format` to allow setting this option from the cmake function later.

mm318 (approved):
Looks good to me, pending CI (I will start CI as soon as I can).

mm318:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13452/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8363/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11176/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13516/)

mikepurvis:
> I updated the documentation to include instructions for setting these variables. I will create a follow-on PR for `ament_cmake_clang_format` to allow setting this option from the cmake function later.

AFAICT this didn't end up happening?

PR #287, opened 2021-01-10
Add --packages-select to ament_clang_tidy

tylerjw:

# Description

This adds a way to filter based on the package name.  This does not add any dependencies and makes this simpler to use for a specific package.  The package names are derived from the directory names that compile_commands.json files are found it which is something already done in this script on line 116.

This will make my use case of this much simpler if I select the packages to run this over.

## Context

This package's REAMDE contains an example of why I want this feature: https://github.com/tylerjw/moveit_ci_tools


ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -91,6 +94,11 @@ def main(argv=sys.argv[1:]):
         start_time = time.time()
 
     compilation_dbs = get_compilation_db_files(args.paths)
+    if args.packages_select:
+        if len(args.packages_select) == 1:
+            args.packages_select = args.packages_select[0].strip().split()
```
tylerjw, line 99: Note that the point of these lines is that if someone has a newline-separated string of package names that they would like to use with this argument it will be handled nicely.  An example of a command that generates a newline-separated list of packages is `colcon list --names-only`.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -77,6 +77,9 @@ def main(argv=sys.argv[1:]):
         '--system-headers',
         action='store_true',
         help='Displays errors from all system headers')
+    parser.add_argument(
+        '--packages-select', nargs='+', metavar='PKG_NAME',
```
mm318, line 81: Can the `metavar` be something like `PKG_NAME1 [PKG_NAME2] ...` (or whatever the convention is)?

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -91,6 +94,11 @@ def main(argv=sys.argv[1:]):
         start_time = time.time()
 
     compilation_dbs = get_compilation_db_files(args.paths)
+    if args.packages_select:
+        if len(args.packages_select) == 1:
```
mm318, line 98: Does this mean `nargs` should be `1` instead of `+`?

mm318 (commented):
Can you also please update the [doc](https://github.com/ament/ament_lint/blob/master/ament_clang_tidy/doc/index.rst)? Thanks!

ament_clang_tidy/ament_clang_tidy/main.py
tylerjw: I copied this from the `colcon` argument `--packages-select` so it has the same form as the `colcon` arg of the same name.

ament_clang_tidy/ament_clang_tidy/main.py
tylerjw: What this does is handle the case where the user passes in a bash variable or a quoted list. In that case, the quoted list is put in the array as a single string with spaces separating it. This if statement breaks up that quoted list of package names into the list of package names we need for this change.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -77,6 +77,9 @@ def main(argv=sys.argv[1:]):
         '--system-headers',
         action='store_true',
         help='Displays errors from all system headers')
+    parser.add_argument(
+        '--packages-select', nargs='*', metavar='PKG_NAME',
+        help='Only process a subset of packages')
```
tylerjw, line 82: This was copied directly from the `colcon` source for the argument of the same name.  I wanted to achieve the same sort of behavior as there is in colcon for selecting packages.

ament_clang_tidy/ament_clang_tidy/main.py
tylerjw: This is needed because if someone passes a newline-separated escaped string of package names it will be a one length list with the string.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -91,14 +98,22 @@ def main(argv=sys.argv[1:]):
...
+        compilation_dbs = filter_packages_select(compilation_dbs, args.packages_select)
+
     if not compilation_dbs:
         print('No compilation database files found', file=sys.stderr)
         return 1
 
     bin_names = [
         # 'clang-tidy',
-        'clang-tidy-6.0',
+        'clang-tidy-' + args.clang_tidy_version,
```
mm318, line 114: I would make this a separate change. Supporting other versions of `clang-tidy` requires enhancements to the parsing of errors and warnings. The regex at https://github.com/ament/ament_lint/blob/34ef6d2b9cb3c11468d7f5181c4fa17269822188/ament_clang_tidy/ament_clang_tidy/main.py#L184 just won't cut it.

tylerjw: Ok, I'll split this change out.  I should have known better.  The reason I did this was that we have standardized on clang-10 tooling for moveit in both ros1 and 2.  As far as I can tell this works with clang-tidy-10.

mm318: The `stdout` output may be fine, but I think for certain kinds of errors and warning returned by `clang-tidy-10`, the xUnit XML file will contain nonsensical snippets.

Either way, I think adding support for a newer version of `clang-tidy` needs a closer look.

tylerjw: I'll split this change out now and do some testing.  Maybe I can figure out a straightforward way to support clang-tidy-10.  It would make using this much easier as 10 is much newer and therefore more easily available in newer Ubuntu now and in the future (hence why we changed in MoveIt).

mm318:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13461/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8372/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11185/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13525/)

Issue #283, opened 2021-01-05
ament_cmake_clang_tidy: No compilation database files found


nnmm:
It's unclear to me how to run clang-tidy in addition to the `ament_lint_common` linter suite. I'm building my package with `--cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=1`, which creates `build/compilation_database.json`, but `ament_cmake_clang_tidy` doesn't seem to pick it up and instead gives the error in the title.

In `CMakeLists.txt`, I have

    find_package(ament_lint_auto REQUIRED)
    ament_lint_auto_find_test_dependencies()

and in `package.xml`

    <test_depend>ament_lint_auto</test_depend>
    <test_depend>ament_cmake_clang_tidy</test_depend>
    <test_depend>ament_lint_common</test_depend>

Is something missing there?

tylerjw:
Based on how it works I believe that error is being caused by it being executed from some working driectory where searching for compile_commands.json is failing.  If you look at the documentation for running it directly (https://github.com/ament/ament_lint/blob/master/ament_clang_tidy/doc/index.rst)  you see there is a positional argument to pass in the path to use for the search for the compile_commands.json files.  I've been looking around for how to pass arguemnts into the cmake ament linters and I've yet to figure it out.

tylerjw:
Based on this line: https://github.com/ament/ament_lint/blob/8bf194aa1ac282db5483dd0d3fefff8f325b0db8/ament_cmake_clang_tidy/cmake/ament_clang_tidy.cmake#L49

It looks like the cmake function `ament_lint_auto_find_test_dependencies` has no way to know that this argument should be passed.  I think if I understand it correctly if you want to use ament_clang_tidy you can't use the auto lint feature.  Instead in your cmake you can do this:

```cmake
find_package(ament_cmake REQUIRED)
if(BUILD_TESTING)
  find_package(ament_clang_tidy REQUIRED)
  ament_clang_tidy(${CMAKE_BINARY_DIR})
endif()
```

I haven't tested this and it is contrary to the documentation for ament_cmake_clang_tidy (https://github.com/ament/ament_lint/blob/master/ament_cmake_clang_tidy/doc/index.rst) as I added the argument `${CMAKE_BINARY_DIR}`.  Maybe someone who understands this better can respond.

tylerjw:
I tested this in my workspace and I needed this to run clang-tidy:

```cmake
find_package(ament_cmake REQUIRED)
[...]

PR #300, opened 2021-02-14
Add ament_cmake_clang_tidy to ament_lint_common

mm318:
Closes https://github.com/ament/ament_lint/issues/283

This pull request will make `ament_clang_tidy` one of the default linters that get run during `colcon test`.

### DIFF
--- a/ament_cmake_clang_tidy/cmake/ament_cmake_clang_tidy_lint_hook.cmake
+++ b/ament_cmake_clang_tidy/cmake/ament_cmake_clang_tidy_lint_hook.cmake
@@ -24,5 +24,5 @@ file(GLOB_RECURSE _source_files FOLLOW_SYMLINKS
 )
 if(_source_files)
   message(STATUS "Added test 'clang_tidy' to check C / C++ code style")
-  ament_clang_tidy()
+  ament_clang_tidy(${CMAKE_BINARY_DIR})
 endif()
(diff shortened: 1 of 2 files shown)


audrow:
Let me know when this is ready for a review.

Do you think that this will add a bunch of warnings downstream? Would it make sense to run CI for all downstream packages?

mwcondino:
Is this still being actively worked? I'd be interested in picking it up if not

mm318:
Sure @mwcondino, any help would be appreciated!

I think this involves more than just turning on `clang-tidy` though, such as assessing the impact on the building of current packages.

nuclearsandwich:
I've been doing some work to get clang-tidy running for a subset of ROS packages. An example build from that work is here: https://build.ros2.org/view/Rci/job/Rci__clang-tidy_ubuntu_focal_amd64/6/

Adding additional tools to ament_lint_common will, in my opinion, constitute a major version level change for the ament_lint packages and require a lot of communication to downstream consumers outside of the ROS 2 core.

And regarding the core, I do not think that clang-tidy can be added to `ament_lint_common` without a massive effort to review and squash existing warnings in the ROS 2 core packages. Some review is happening now and so we can expect some improvement from the state listed in the linked job but I don't think making clang_tidy part of `ament_lint_common` is likely during the ROS Humble development phase.

However, one of the conflicts this PR currently has is due to https://github.com/ament/ament_lint/pull/316 which includes the same change in the CMake hook to properly set the script in the CMake build directory. I still have to work with @audrow to coordinate a release of ament_lint but if you're already using `ament_lint_common` and `ament_lint_auto`, you can add `<test_depend>ament_cmake_clang_tidy</test_depend>` and `ament_lint_auto` should find and hook it just like the linters in ament_lint_common.

audrow:
Closing because it has been years. Feel free to reopen.

PR #315, opened 2021-06-17
Add SPDX identifiers to the licenses.

clalancette:
This is so we can use them at a higher layer.  Note that the
SPDX identifier for "BSD License 2.0" is made up, since that
license is also made up.  We eventually need to remove it,
but some of the core packages still depend on it (and really
should be switched over to BSD-3-Clause).

### DIFF
--- a/ament_copyright/ament_copyright/licenses.py
+++ b/ament_copyright/ament_copyright/licenses.py
@@ -51,6 +51,8 @@ def read_license_data(path, name, spdx, prefix):
     return LicenseEntryPoint(name, spdx, file_headers, license_files, contributing_files)
 
 
+# The SPDX identifier (the 3rd argument) comes from the official list at https://spdx.org/licenses/
+
 apache2 = read_license_data(TEMPLATE_DIRECTORY,
                             'Apache License, Version 2.0',
                             'Apache-2.0',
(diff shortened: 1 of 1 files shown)


gbalke:
Could be worth adding a comment mentioning these should be SPDX identifiers? Would be useful to be explicit in case people want to add new licenses!

audrow (approved):
Looks good to me.

clalancette:
> Could be worth adding a comment mentioning these should be SPDX identifiers? Would be useful to be explicit in case people want to add new licenses!

The variable it gets put into is named 'spdx', which seems pretty clear to me.  But maybe it is worthwhile to point to the URL where I got these; I'll add that.

clalancette:
Full CI (since basically everything depends on this package):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14769/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9518/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12435/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14916/)

clalancette:
The only failures are in mypy, which is a known issue right now.  So this is ready for review.

clalancette:
Approved, and no new test failures.  Going ahead and merging.

PR #316, opened 2021-08-04
Improvements to ament_lint_clang_tidy.

nuclearsandwich:
Without this default in the hook script ament_clang_tidy does not execute properly.

This is still in draft while I determine why a default is needed in CMake when the python script wrapping clang-tidy should be handling the default instead.

The second commit updates the script to check for more than just clang-tidy-6.0.

### DIFF
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -112,7 +112,9 @@ def main(argv=sys.argv[1:]):
         return 1
 
     bin_names = [
-        # 'clang-tidy',
+        'clang-tidy',
+        'clang-tidy-10',
+        'clang-tidy-11',
         'clang-tidy-6.0',
     ]
     clang_tidy_bin = find_executable(bin_names)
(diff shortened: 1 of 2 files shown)


audrow (approved):
This looks okay to me.

nuclearsandwich:
Okay so I spent some time refreshing myself on these changes. I have some reservations about the hook script having a default argument if the default instead ought to be pushed further up, either in the ament_clang_tidy() cmake function or in the ament_clang_tidy script. ament_clang_tidy defaults to looking in `os.curdir` and when run via ament_cmake_clang_tidy the working directory is the source directory. In this default state for an out-of-tree build it will fail since the compile commands database file required by clang-tidy is in the build directory rather than the source directory. But in a local test where I changed the working directory to the build directory it fails to locate source files properly.

I am less concerned with the hook containing a default argument after scrolling through some of the other hook scripts. Both ament_cppcheck and ament_cpplint have default variables.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15135/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9864/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12823/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15307/)

ament_cmake_clang_tidy isn't enabled by default in ament_lint_common so won't be used by default in any ROS 2 packages, but I wanted to make sure that the packages still build and that tests still run coherently with these changes, especially on non-linux platforms. So this CI run runs test_rclcpp tests along with building and testing ament_cmake_clang_tidy.

PR #319, opened 2021-09-02
ament_uncrustify: update config for uncrustify 0.73

mjbogusz:
Updated the config with:
```sh
uncrustify -c ament_code_style.cfg --update-config-with-doc -o ament_code_style.cfg
```
and updated few configuration options that caused problems here: https://ci.ros2.org/job/ci_linux/13210/

It's still reporting quite a few style problems, but some of them are due to the old uncrustify being inconsistent - e.g. it was double-indenting lambdas, which is not possible with the new version unless I'd switch all the indentation options to `ignore`.

Related to https://github.com/ament/uncrustify_vendor/pull/14 (I'm not sure what the order should be)


clalancette:
This was actually superseded by the work we did in #475, so closing this out.

PR #320, opened 2021-09-02
Enable MISRA checking in cppcheck

mjeronimo:
For Space ROS, the --misra option to cppcheck is enabled by default from ament_cppcheck, since MISRA checking will be required, and a command-line option introduced to disable it.

### DIFF
--- a/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
@@ -81,6 +81,6 @@ if(_source_files)
     STATUS "Configured cppcheck exclude dirs and/or files: ${_all_exclude}"
   )
   ament_cppcheck(
-    ${_language} INCLUDE_DIRS ${_all_include_dirs} EXCLUDE ${_all_exclude}
+    MISRA ${_language} INCLUDE_DIRS ${_all_include_dirs} EXCLUDE ${_all_exclude}
   )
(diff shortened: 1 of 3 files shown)


audrow (approved):
Looks good to me with green CI. It'd probably be good to run CI on all downstream packages.

nuclearsandwich (changes requested):
During my first review pass looking at the overall PR diff. Something about this struck me as a little off. The MISRA option is passed from cmake to ament_cppcheck to cppcheck itself but goes through a weird inversion of expression. In the CMake function MISRA is a "positive" argument. ON means MISRA, OFF means not MISRA. But in ament_cppcheck it's a "negative" argument `--disable-misra` present means not MISRA and absent means MISRA. In cppcheck, the option `--addon=misra` argument is again positive.

I took a look at the commit history and saw 5dd935f110f6de2113e03f8cc6ae2245085138d2 which has all "positive" arguments throughout and I think that's actually the best interface for the MISRA argument.

In order to get MISRA checking on by default for spaceros I would suggest updating the ament_cmake_cppcheck hook script to set `MISRA ON` there. https://github.com/ament/ament_lint/blob/master/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake#L83-L85

mjeronimo:
Yes, I agree. I'll update accordingly.

nuclearsandwich (approved):
Looks good to me once the DCO bot is happy. Thanks for iterating!

PR #337, opened 2021-11-01
remove google style from clang-tidy default settings

wjwwood:
This also removes the need for default config file.

I propose this change because:

- `cpplint.py` covers these Google style-like checks (things like using explicit on a constructor with one argument)
- these are often false positives, and need to be suppressed, but the suppression mechanism (`// NOLINT(rule)`) is problematic because the name of the rules are different for `cpplint.py` and clang-tidy's version of the rules
- when providing multiple rules clang-tidy emits warnings that it doesn't know the rule name used by `cpplint.py`, causing more unhelpful warnings

The simplest solution, in my opinion, is to remove this from the default settings, rely on `cpplint.py` for these kinds of checks, and let individual users add the Google style checks to their on clang-tidy config files, if they wish to use them.

@nuclearsandwich FYI

### DIFF
--- a/ament_clang_tidy/ament_clang_tidy/configuration/.clang-tidy
+++ b/ament_clang_tidy/ament_clang_tidy/configuration/.clang-tidy
@@ -1,2 +1,0 @@
-Checks: >
-    google-*,
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -32,8 +32,6 @@ import yaml
 
 
 def main(argv=sys.argv[1:]):
-    config_file = os.path.join(
-        os.path.dirname(__file__), 'configuration', '.clang-tidy')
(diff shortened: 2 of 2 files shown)


audrow (approved):
This looks good to me.

nuclearsandwich (approved):
Thanks @wjwwood!

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15551/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10235/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13218/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15765/)

nyxrobotics:
Hello, nice to meet you. I have one question: Based on your explanation, it seems like configuring clang-tidy would be the better approach. Could you clarify why you prefer to use cpplint instead? Is this part of a milestone where you plan to eventually transition to clang-tidy in the future?

PR #339, opened 2021-11-11
ignore NOLINT comments with categories that come from clang-tidy

wjwwood:
So, I'm still running into issues with clang-tidy and cpplint colliding when using `NOLINT` comments. This is an idea on how to handle it and I've made an upstream issue to see what they think as well: https://github.com/cpplint/cpplint/issues/184

@nuclearsandwich @mjeronimo FYI

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -279,6 +279,12 @@ _LEGACY_ERROR_CATEGORIES = [
     'readability/function',
     ]
 
+# These prefixes for categories should be ignored since they relate to other
+# tools which also use the NOLINT syntax, e.g. clang-tidy.
+_OTHER_NOLINT_CATEGORY_PREFIXES = [
+    'clang-analyzer',
+    ]
+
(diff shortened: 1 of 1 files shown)


mjeronimo:
Looks like a good idea to me. Also like the idea of pushing it upstream to the tool as well.

wjwwood:
Moving this into review, I would appreciate feedback.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16329/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10949/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16701/)

nuclearsandwich (approved):
No objections to the implementation. It'll be great if this becomes part of our upstream fork.

mjeronimo:
Looks good to me as well.

mjeronimo (approved):

wjwwood:
The test failures look unrelated to me and were not present on Windows. Thanks for the reviews guys. I'll see about following up on our fork if this ends up in a future version of upstream, which seems likely.

PR #334, opened 2021-10-25
[ament_uncrustify] Fix file exclusion behavior

aprotyas:
This PR fixes the file exclusion behavior reported in #326.

Specifically, the exclusion list is matched against
files/directories as the search path is traversed.

Tries to maintain consistency with #327.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

PR #342, opened 2021-11-30
Fix language set to Ament uncrustify

jparisu:
Ament uncrustify did not use correctly the `--language` parameter. If the extension of the file was different than a C or C++ standard extension (`c_extensions = ['c', 'cc', 'h', 'hh'] ; cpp_extensions = ['cpp', 'cxx', 'hpp', 'hxx']`) the file was parsed by default by uncrustify instead of `--language` specific option.

It also adds `ipp` as a C++ extension (implementation file) used, for example, for template implementations.

### DIFF
--- a/ament_uncrustify/ament_uncrustify/main.py
+++ b/ament_uncrustify/ament_uncrustify/main.py
@@ -37,7 +37,7 @@ def main(argv=sys.argv[1:]):
         'configuration', 'ament_code_style.cfg')
 
     c_extensions = ['c', 'cc', 'h', 'hh']
-    cpp_extensions = ['cpp', 'cxx', 'hpp', 'hxx']
+    cpp_extensions = ['cpp', 'cxx', 'hpp', 'hxx', 'ipp']
 
     parser = argparse.ArgumentParser(
         description='Check code style using uncrustify.',
@@ -243,18 +243,32 @@ def get_files(paths, extension_types, exclude_patterns=[], language=None):
                 # select files by extension
                 for filename in sorted(filenames):
                     _, ext = os.path.splitext(filename)
-                    language = extensions_with_dot_to_language.get(ext, None)
-                    if language is not None:
-                        filepath = os.path.join(dirpath, filename)
-                        if os.path.realpath(filepath) not in excludes:
+                    # Check whether to use default ext language or user custom one
+                    language_ = extensions_with_dot_to_language.get(ext, None)
+                    filepath = os.path.join(dirpath, filename)
+                    if os.path.realpath(filepath) not in excludes:
+                        # Add file only if it is not excluded
+                        if language_ is not None:
+                            # Add file with its own extension
+                            files[language_].append(
+                                os.path.normpath(os.path.join(dirpath, filename)))
+                        else:
+                            # When ext not found, add file with user custom extension
                             files[language].append(
                                 os.path.normpath(os.path.join(dirpath, filename)))
+
         if os.path.isfile(path):
             _, ext = os.path.splitext(path)
-            language = extensions_with_dot_to_language.get(ext, None)
-            if language is not None:
-                if os.path.realpath(path) not in excludes:
+            # Check whether to use default ext language or user custom one
+            language_ = extensions_with_dot_to_language.get(ext, None)
+            if os.path.realpath(path) not in excludes:
+                if language_ is not None:
+                    # Add file with its own extension
+                    files[language_].append(os.path.normpath(path))
+                else:
+                    # When ext not found, add file with user set extension
                     files[language].append(os.path.normpath(path))
+
     return files
 
 


aprotyas:
@jparisu thank you for the PR! It looks like you accidentally committed your build artifacts as well. Do you mind amending the commits to exclude your build directory?

You can do so as follows from the base of your repository:

```bash
git reset --soft HEAD~2
git reset HEAD build
git commit -c ORIG_HEAD
```

aprotyas (commented):
@jparisu these changes LGTM, but I'm introducing almost identical changes (bar support for IPP extensions) to fix the "default" language issue in #334. Do you mind waiting for that to merge so you can rebase and add the IPP extensions? That way there will be least duplication of efforts.

jparisu:
Hi @aprotyas , thank you for your fast answer.

> Do you mind waiting for that to merge so you can rebase and add the IPP extensions? That way there will be least duplication of efforts.

I do not mind at all. We will use my branch for the moment (we use it in some github actions).

clalancette:
@jparisu Now that #334 has merged, would you mind rebasing on top of that?

jparisu:
> @jparisu Now that #334 has merged, would you mind rebasing on top of that?

There you go. :+1:

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -243,18 +243,32 @@ def get_files(paths, extension_types, exclude_patterns=[], language=None):
...
+                    if os.path.realpath(filepath) not in excludes:
+                        # Add file only if it is not excluded
+                        if language_ is not None:
+                            # Add file with its own extension
+                            files[language_].append(
+                                os.path.normpath(os.path.join(dirpath, filename)))
+                        else:
+                            # When ext not found, add file with user custom extension
                             files[language].append(
                                 os.path.normpath(os.path.join(dirpath, filename)))
```
aprotyas, line 258: Is this desired behavior? This will let the tool pick up files that are (probably) not C/C++ files.

```bash
/home/dev/test# ls
test.cpp  test.foo  test.ipp  test.launch.py  test.xml

/home/dev/test# ament_uncrustify . --language C++
Output prefix: /tmp/uncrustify_x5mupoa_/
Output suffix: .uncrustify
Parsing: /home/dev/test.cpp as language CPP
Parsing: /home/dev/test.foo as language CPP
Parsing: /home/dev/test.ipp as language CPP
Parsing: /home/dev/test.launch.py as language CPP
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 2, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 6, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 8, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 24, parent is NONE: Unmatched CONSTR_COLON
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 33, parent is NONE: Unmatched ASSIGN
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 35, parent is NONE: Unmatched RETURN
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 2, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 6, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 8, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 24, parent is NONE: Unmatched CONSTR_COLON
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 33, parent is NONE: Unmatched ASSIGN
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 35, parent is NONE: Unmatched RETURN
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 2, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 6, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 8, parent is NONE: Unmatched NEWLINE
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 24, parent is NONE: Unmatched CONSTR_COLON
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 33, parent is NONE: Unmatched ASSIGN
indent_text(3155): size is 7
indent_text(3157): File: /home/dev/test.launch.py, open_line is 35, parent is NONE: Unmatched RETURN
Parsing: /home/dev/test.xml as language CPP

The invocation of 'uncrustify' failed with error code 1: Command '['/usr/bin/uncrustify', '-c', '/home/dev/workspaces/ament_ws/build/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg', '-l', 'CPP', '--prefix', '/tmp/uncrustify_x5mupoa_', '--suffix', '.uncrustify', '/home/dev/test.cpp', '/home/dev/test.foo', '/home/dev/test.ipp', '/home/dev/test.launch.py', '/home/dev/test.xml']' returned non-zero exit status 1
```

FYI, in this example, `test.xml` and `test.launch.py` were in fact a pair of well-formed XML file and launch script (and not C/C++ sources). This causes `uncrustify` to choke when linting.

---

I think if we want to allow "user set extension" support, we need to add an argument to the `ament_uncrustify` CLI that will let users add to the supported extension set at runtime.

ament_uncrustify/ament_uncrustify/main.py
```diff
@@ -243,18 +243,32 @@ def get_files(paths, extension_types, exclude_patterns=[], language=None):
...
+            # Check whether to use default ext language or user custom one
+            language_ = extensions_with_dot_to_language.get(ext, None)
+            if os.path.realpath(path) not in excludes:
+                if language_ is not None:
+                    # Add file with its own extension
+                    files[language_].append(os.path.normpath(path))
+                else:
+                    # When ext not found, add file with user set extension
                     files[language].append(os.path.normpath(path))
+
```
aprotyas, line 271: Same comment as above.

aprotyas (changes requested):
I think we can limit the scope of this PR to increasing the default extension set to match against, but I would advise against the changes proposed to the file matching mechanism. Check a sample run inline to see why. Thoughts?

ament_uncrustify/ament_uncrustify/main.py
jparisu: I think the problem here is the help test of this tool.

`ament_uncrustify -h` :
```
  --language {C,C++,CPP}
                        Passed to uncrustify as '-l <language>' to force a specific language rather then choosing one based on file extension (default: None)
```
I understood that was the desired behaviour, so fyes, this is the expected behaviour. I think that when you set `--language` argument, it is because you want a specific file to be parsed as a specific format.

Indeed, I think this argument should not be used whit multiple files, but this is a user's problem (as long as it is correctly specified in the program help).

Thus, please @aprotyas , tell me if you approve, or if any change should be done to the help description.

aprotyas: > Indeed, I think this argument should not be used whit multiple files, but this is a user's problem (as long as it is correctly specified in the program help).

@jparisu I see your point, but I think it's unreasonable to expect the user to manually exclude all files they don't want uncrustify to look at. For example, the example I showed above would need to be padded with a bunch of excludes if we shifted this responsibility to users.

I would suggest reverting the "match arbitrary extension" behavior. Instead, I think a better option could be to still specify a language, but allow the user to add to the set of extensions being matched against at runtime. Say:

`ament_uncrustify . --language C++ --extensions .cc .cp .c++`

The `--extensions` option would take a list of arguments that would be added to the list of extensions against which files are matched.

jparisu: I am sorry but I just wanted to fix a bug that was in the code and/or in the manual of the executable. 
Actually, I just wanted my `.ipp` files to be parsed as C++ :sweat_smile: .

This seems to me like changing the behaviour of the execution, or like adding a new feature. And I do not think I am the correct one to do it.

Thus, we will remain using my fork to parse `.ipp` files for now.
If there is something else I could do, let me know.

PR #345, opened 2022-01-11
Disable cppcheck 2.x.

clalancette:
It has known performance issues (more than 10x slower than
cppcheck 1.9), so just get out early and don't run it.

### DIFF
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -122,8 +122,8 @@ def main(argv=sys.argv[1:]):
         # the number of cores cannot be determined, do not extend args
         pass
 
-    # detect cppcheck 1.88 which caused issues
-    if 'AMENT_CPPCHECK_ALLOW_1_88' not in os.environ:
+    # detect cppcheck 1.88 or 2.x which are much too slow
+    if 'AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS' not in os.environ:
         if cppcheck_version == '1.88' or cppcheck_version.startswith('2.'):
             print(
                 f'cppcheck {cppcheck_version} has known performance issues and therefore will not '
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -39,11 +39,15 @@ def find_cppcheck_executable():
 def get_cppcheck_version(cppcheck_bin):
     version_cmd = [cppcheck_bin, '--version']
     output = subprocess.check_output(version_cmd)
-    # expecting something like b'Cppcheck 1.88\n'
+    # expecting something like b'Cppcheck 1.88\n' or b'Cppcheck 2.7 dev'
     output = output.decode().strip()
     tokens = output.split()
-    if len(tokens) != 2:
+    if len(tokens) not in (2, 3):
         raise RuntimeError("unexpected cppcheck version string '{}'".format(output))
+
+    if tokens[0] != 'Cppcheck':
+        raise RuntimeError("unexpected cppcheck version name '{}'".format(output))
+
     return tokens[1]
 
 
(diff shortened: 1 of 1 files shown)


ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -120,18 +124,18 @@ def main(argv=sys.argv[1:]):
 
     # detect cppcheck 1.88 which caused issues
     if 'AMENT_CPPCHECK_ALLOW_1_88' not in os.environ:
```
aprotyas, line 126: In light of the version check happening below, this environment variable seems like a misnomer, though I don't have an actionable recommendation here.

clalancette: Ah yeah, thanks for bringing it up.

I considered changing the name of the environment variable, but I didn't bother.  Mostly I wanted to keep backwards compatibility, and I thought adding a new environment variable was overkill.  Also I was being lazy.

But I think the chances that someone was depending on `AMENT_CPPCHECK_ALLOW_1_88` is really small, so I could be convinced to change the name.

aprotyas: I can't think of a reasonable variable name to replace with, but `AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS` is the best I have.

Alternatively the `1.88` and `2.x` versions can be allowed/checked separately, so there would be a pair of environment variables: `AMENT_CPPCHECK_ALLOW_1_88`, `AMENT_CPPCHECK_ALLOW_2_X`. I'm not sure if this option is worth the hassle.

audrow (approved):
Looks good to me with green CI. I'm okay with the very specific environment variable.

ament_cppcheck/ament_cppcheck/main.py
clalancette: I ended up taking your suggestion of `AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS` in 16ad69d.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10626/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16275/)

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -118,20 +122,20 @@ def main(argv=sys.argv[1:]):
...
-    if 'AMENT_CPPCHECK_ALLOW_1_88' not in os.environ:
-        if cppcheck_version == '1.88':
+    # detect cppcheck 1.88 or 2.x which are much too slow
+    if 'AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS' not in os.environ:
+        if cppcheck_version == '1.88' or cppcheck_version.startswith('2.'):
             print(
-                'cppcheck 1.88 has known performance issues and therefore will not be used, '
-                'set the AMENT_CPPCHECK_ALLOW_1_88 environment variable to override this.',
+                f'cppcheck {cppcheck_version} has known performance issues and therefore will not '
+                'be used, set the AMENT_CPPCHECK_ALLOW_1_88 environment variable to override this.',
```
aprotyas, line 130:
```suggestion
                'be used, set the AMENT_CPPCHECK_ALLOW_SLOW_VERSIONS environment variable to override this.',
```

aprotyas (approved):
Small nit inline but approving regardless.

ament_cppcheck/ament_cppcheck/main.py
clalancette: Bah, good catch.  Fixed in 58ad06d.  I tested this out locally and it seems to be happy.  I'm not going to run another full CI on this; once the previous CI is done I'll go ahead and merge.

PR #368, opened 2022-04-08
Add SARIF output to ament_clang_tidy

mjeronimo:
* Add a --sarif-file command-line option
* Convert from the internal report format to SARIF
* Write the SARIF file

Fixes: https://github.com/space-ros/space-ros/issues/17

### DIFF
--- a/ament_clang_tidy/ament_clang_tidy/main.py
+++ b/ament_clang_tidy/ament_clang_tidy/main.py
@@ -89,6 +89,9 @@ def main(argv=sys.argv[1:]):
     parser.add_argument(
         '--xunit-file',
         help='Generate a xunit compliant XML file')
+    parser.add_argument(
+        '--sarif-file',
+        help='Generate a SARIF file')
     args = parser.parse_args(argv)
 
     if not os.path.exists(args.config_file):
@@ -96,9 +99,7 @@ def main(argv=sys.argv[1:]):
               file=sys.stderr)
         return 1
 
-    if args.xunit_file:
-        start_time = time.time()
-
+    start_time = time.time()
     compilation_dbs = get_compilation_db_files(args.paths)
 
     if args.packages_select is not None:
@@ -242,6 +243,8 @@ def main(argv=sys.argv[1:]):
         if current_file is not None:
             report[current_file].append(copy.deepcopy(data))
 
+    elapsed_time = time.time() - start_time
+
     if args.xunit_file:
         folder_name = os.path.basename(os.path.dirname(args.xunit_file))
         file_name = os.path.basename(args.xunit_file)
@@ -252,12 +255,20 @@ def main(argv=sys.argv[1:]):
             if file_name.endswith(suffix):
                 file_name = file_name[0:-len(suffix)]
         testname = '%s.%s' % (folder_name, file_name)
-        xml = get_xunit_content(report, testname, time.time() - start_time)
+        xml = get_xunit_content(report, testname, elapsed_time)
         path = os.path.dirname(os.path.abspath(args.xunit_file))
         if not os.path.exists(path):
             os.makedirs(path)
         with open(args.xunit_file, 'w') as f:
             f.write(xml)
+
+    if args.sarif_file:
+        sarif = get_sarif_content(report, get_clang_tidy_version(clang_tidy_bin))
+        path = os.path.dirname(os.path.abspath(args.sarif_file))
+        if not os.path.exists(path):
+            os.makedirs(path)
+        with open(args.sarif_file, 'w') as f:
+            f.write(sarif)
 
 
 def find_executable(file_names):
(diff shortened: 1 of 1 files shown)


mjeronimo:
An example output file, for reference:

```json
{
  "version": "2.1.0",
  "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.5",
  "properties": {
    "comment": "clang-tidy output converted to SARIF by ament_clang_tidy"
  },
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "clang-tidy",
          "version": "10.0.0",
          "informationUri": "https://clang.llvm.org/extra/clang-tidy/",
          "rules": [
            {
              "id": "google-explicit-constructor",
              "shortDescription": {
                "text": "single-argument constructors must be marked explicit to avoid unintentional implicit conversions"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            },
            {
              "id": "google-explicit-constructor",
              "shortDescription": {
                "text": "single-argument constructors must be marked explicit to avoid unintentional implicit conversions"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            },
            {
              "id": "clang-analyzer-core.CallAndMessage",
              "shortDescription": {
                "text": "1st function call argument is an uninitialized value"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            },
            {
              "id": "clang-diagnostic-uninitialized",
              "shortDescription": {
                "text": "variable 'some_value' is uninitialized when used here"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            },
            {
              "id": "clang-analyzer-core.CallAndMessage",
              "shortDescription": {
                "text": "1st function call argument is an uninitialized value"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            },
            {
              "id": "clang-diagnostic-uninitialized",
              "shortDescription": {
                "text": "variable 'x' is uninitialized when used here"
              },
              "helpUri": "https://clang.llvm.org/extra/clang-tidy/checks/list.html"
            }
          ]
        }
      },
      "artifacts": [
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/include/rcpputils/filesystem_helper.hpp"
          }
        },
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/src/asserts.cpp"
          }
        },
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/src/env.cpp"
          }
        },
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/src/filesystem_helper.cpp"
          }
        },
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/src/find_library.cpp"
          }
        },
        {
          "location": {
            "uri": "/tmp/spaceros/src/rcpputils/src/shared_library.cpp"
          }
        }
      ],
      "results": [
        {
          "ruleId": "google-explicit-constructor",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "  path(const std::string & p);  // NOLINT(runtime/explicit): this is a conversion constructor\n  ^\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/include/rcpputils/filesystem_helper.hpp",
                  "index": 0
                },
                "region": {
                  "startLine": "89",
                  "startColumn": "3"
                }
              }
            }
          ]
        },
        {
          "ruleId": "google-explicit-constructor",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "  path(const std::string & p);  // NOLINT(runtime/explicit): this is a conversion constructor\n  ^\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/include/rcpputils/filesystem_helper.hpp",
                  "index": 0
                },
                "region": {
                  "startLine": "89",
                  "startColumn": "3"
                }
              }
            }
          ]
        },
        {
          "ruleId": "clang-analyzer-core.CallAndMessage",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "  jnk(some_value);\n  ^\n/tmp/spaceros/src/rcpputils/src/asserts.cpp:23:3: note: 'some_value' declared without an initial value\n  int some_value;\n  ^\n/tmp/spaceros/src/rcpputils/src/asserts.cpp:24:3: note: 1st function call argument is an uninitialized value\n  jnk(some_value);\n  ^\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/src/asserts.cpp",
                  "index": 1
                },
                "region": {
                  "startLine": "24",
                  "startColumn": "3"
                }
              }
            }
          ]
        },
        {
          "ruleId": "clang-diagnostic-uninitialized",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "  jnk(some_value);\n      ^\n/tmp/spaceros/src/rcpputils/src/asserts.cpp:23:17: note: initialize the variable 'some_value' to silence this warning\n  int some_value;\n                ^\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/src/asserts.cpp",
                  "index": 1
                },
                "region": {
                  "startLine": "24",
                  "startColumn": "7"
                }
              }
            }
          ]
        },
        {
          "ruleId": "clang-analyzer-core.CallAndMessage",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "    foobar(x);\n    ^\n/tmp/spaceros/src/rcpputils/src/env.cpp:53:3: note: 'x' declared without an initial value\n  int x;\n  ^\n/tmp/spaceros/src/rcpputils/src/env.cpp:54:7: note: Assuming the condition is true\n  if (!rcutils_set_env(env_var, env_value)) {\n      ^\n/tmp/spaceros/src/rcpputils/src/env.cpp:54:3: note: Taking true branch\n  if (!rcutils_set_env(env_var, env_value)) {\n  ^\n/tmp/spaceros/src/rcpputils/src/env.cpp:55:5: note: 1st function call argument is an uninitialized value\n    foobar(x);\n    ^\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/src/env.cpp",
                  "index": 2
                },
                "region": {
                  "startLine": "55",
                  "startColumn": "5"
                }
              }
            }
          ]
        },
        {
          "ruleId": "clang-diagnostic-uninitialized",
          "level": "warning",
          "kind": "review",
          "message": {
            "text": "    foobar(x);\n           ^\n/tmp/spaceros/src/rcpputils/src/env.cpp:53:8: note: initialize the variable 'x' to silence this warning\n  int x;\n       ^\n        = 0\n"
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/tmp/spaceros/src/rcpputils/src/env.cpp",
                  "index": 2
                },
                "region": {
                  "startLine": "55",
                  "startColumn": "12"
                }
              }
            }
          ]
        }
      ]
    }
  ]
}
```

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -252,13 +255,21 @@ def start_subprocess(full_cmd):
             if file_name.endswith(suffix):
                 file_name = file_name[0:-len(suffix)]
         testname = '%s.%s' % (folder_name, file_name)
-        xml = get_xunit_content(report, testname, time.time() - start_time)
+        xml = get_xunit_content(report, testname, elapsed_time)
```
wjwwood, line 258: nitpick: The changes to capture elapsed time are fine, but they seem orthogonal to the pull request (supporting a sarif output).

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -377,5 +388,105 @@ def get_xunit_content(report, testname, elapsed):
...
+def get_clang_tidy_version(clang_tidy_bin):
+    version = ''
+    try:
+        cmd_args = [clang_tidy_bin, '--version']
+        output = subprocess.check_output(cmd_args).decode()
+        m = re.search('.*LLVM version (.*)\n', output)
+        if m:
+            version = m.group(1)
+    except Exception as e:
+        print(f'Failed to get the clang tidy version number: {e}', file=sys.stderr)
```
wjwwood, line 400:
```suggestion
        print(f"Failed to get the version from '{' '.join(cmd_args)}': {e}", file=sys.stderr)
```

Should this throw? Is it ok to just return `''` when the version cannot be determined? Also, you end up returning `''` if the regex doesn't match, so same question in that case.

wjwwood (commented):
These new functions seem like they could benefit from some simple unit tests. It's hard for me (not knowing the SARIF schema very well) to know if there are any bad assumptions in the logic. Some tests might help with that too.

That being said, what I can understand of it lgtm.

I also had some questions about behavior in certain failure cases, but those aren'y blocking.

ament_clang_tidy/ament_clang_tidy/main.py
mjeronimo: That was a vestige of the refactoring. During development, the get_sarif_content had the same signature as get_xunit_content, requiring the elapsed time to be passed to both. Later, unused params were dropped from the get_sarif_content interface. 

So, I agree that the result seems to be orthogonal, I think it is a relevant change as the elapsed time isn't specific to the xunit format (and shouldn't be be in that code block. Instead, the elapsed time should be computed before any file output occurs.

ament_clang_tidy/ament_clang_tidy/main.py
mjeronimo: In the unlikely case that we can't get the clang tidy version number, I think it's OK to proceed. I've updated the code to return 'unknown' instead of the empty string and warn if either the regex fails or the clang_tidy invocation fails.

wjwwood (approved):

PR #371, opened 2022-05-03
recommend use of --mixin compile-commands

wjwwood:
Kind of a follow up of https://github.com/ament/ament_lint/pull/287, as the `--packages-select` option only makes sense with a combined `compile_commands.json` in the build folder, which you only get with this mixin because it aggregates them for you there. Otherwise each package would generate its own `compile_commands.json` file.

### DIFF
--- a/ament_clang_tidy/doc/index.rst
+++ b/ament_clang_tidy/doc/index.rst
@@ -12,7 +12,7 @@ How to run the check from the command line?
 
 *Prerequisites*: ``clang-tidy-6.0``, ``clang-tools-6.0``, and ``python-yaml`` packages should
 have already been installed. ``compile_commands.json`` files should have already been generated
-(e.g.: workspace built with ``colcon build --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON``).
+(e.g.: workspace built with ``colcon build --mixin compile-commands``).
 
 .. code:: sh
 


jacobperron:
It's strange to me to see references to colcon here to begin with, but presumably our typical workflow for using ament_clang_tidy involves colcon. Maybe it deserves it's own special section in the documentation, e.g. "Using colcon to generate compile_commands.json". I think we could use the section to clear up some assumptions such as how to install colcon mixin support and the `compile-commands` mixin (or it appears it's a default mixin).

wjwwood:
True, but yes, typically colcon is being used. Also that was the case before (mentioning colcon), this pr just updates it to suggest using the mixin which, AFAIK, is the recommended way to generate the compile commands.

This is just a hint in these docs, an "e.g.", which to me implies there might be other ways to do it. But these "prerequisites" are really left up to the user to ensure, whether they are using colcon or not. The point of this document is how to run it on the command line assuming you've managed the prerequisites.

> Maybe it deserves it's own special section in the documentation, e.g. "Using colcon to generate compile_commands.json".

Maybe, but that's more than I was hoping to accomplish with this pr.

---

Also, I was wrong about the:

> only makes sense with a combined compile_commands.json in the build folder,

part, because it actually works even if the compile commands files are distributed in the build folders. So it works as-is, but I still think using the mixin should be the recommended way.

jacobperron (approved):

PR #380, opened 2022-05-07
Enable SARIF output for Space ROS testing

mjeronimo:
Add the --sarif-file command-line option to the invocation of all of the static analysis tools (via colcon test).

Also, ament_copyright requires some special handling because it has some mutually exclusive options, but --sarif-file and --xunit-file should both be allowed during the same run. 

Fixes: https://github.com/space-ros/space-ros/issues/41

### DIFF
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -100,7 +100,7 @@ def main(argv=sys.argv[1:]):
     # such that --xunit-file and --sarif-file options can be used together, but are still part of
     # the mutually exclusive group. For example: XOR(A, B, C, D, (E OR F)).
     if any([
-            args.add_missing, args.add_copyright_year != None,
+            args.add_missing, args.add_copyright_year is not None,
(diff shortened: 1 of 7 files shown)


ament_copyright/ament_copyright/main.py
```diff
@@ -91,11 +91,36 @@ def main(argv=sys.argv[1:]):
...
         help='Generate a xunit compliant XML file')
-    group.add_argument(
+    parser.add_argument(
         '--sarif-file',
         help='Generate a SARIF file')
     args = parser.parse_args(argv)
 
+    # Have to do some additional argument testing here because argparse doesn't support nested groups,
+    # such that --xunit-file and --sarif-file options can be used together, but are still part of
+    # the mutually exclusive group. For example: XOR(A, B, C, D, (E OR F)).
```
wjwwood, line 101: That sucks but I don't see a better solution after looking briefly.

wjwwood (approved):

PR #386, opened 2022-05-30
[ament_lint_auto] General file exclusion with AMENT_LINT_AUTO_FILE_EXCLUDE

### LINKED ISSUE #343: [ament_lint_auto] General file exclusion lists
## What?

A mechanism to provide a list of files to be excluded to `ament_lint_auto`, with the intention being that all linters running as a result of having invoked `ament_lint_auto` will inherit these exclusions.

## Why?

The main utility with this mechanism would be in declaring a list of common C++/Python source files that should not be linted within a package - common examples would be third-party files or templates.

In working on https://github.com/ros2/geometry2/pull/469, I had to deal with the former type of files mentioned above. I noticed that it would be convenient if we had a way to provide a general file exclusion list to `ament_lint_auto`. It would simplify this workflow:

<details>
<summary>Current CMakeLists.txt snippet</summary>

```cmake
# Exclusion list contains ${CMAKE_SOURCE_DIR}/foo
set(_linter_excludes foo)
find_package(ament_lint_auto REQUIRED)
list(APPEND AMENT_LINT_AUTO_EXCLUDE
  ament_cmake_cppcheck
  ament_cmake_cpplint
  ament_cmake_uncrustify
)
ament_cppcheck(EXCLUDE ${_linter_excludes})
ament_cpplint(EXCLUDE ${_linter_excludes})
ament_uncrustify(EXCLUDE ${_linter_excludes})
```

</details>

To the following workflow:

<details>
<summary>Desired CMakeLists.txt snippet</summary>

```cmake
find_package(ament_lint_auto REQUIRED)
# Linters will skip ${CMAKE_SOURCE_DIR}/foo
list(APPEND AMENT_LINT_AUTO_FILE_EXCLUDE
  foo
)
```

</details>

The pain point in the former being the need to manually exclude each of the linters that's supposed to receive the *same* file exclusion list -- if so, why not just signal that file exclusion list to the linter aggregator?

## Implementation consideration

Introducing the ability to populate/append a list variable such as `AMENT_LINT_AUTO_FILE_EXCLUDE` with the general file exclusion list is probably the path with least friction. Having said that, I have not fully thought through the implementation or if there are any ways a general file exclusion list like this can bite.

Thoughts?

aprotyas:
Currently, there is no interface to specify a file exclusion list for linters invoked through ament_lint_auto, and as such each linter needs to be provided with duplicates of a common file exclusion list.

As discussed in #343, it is reasonable to expect that linters registering themselves as ament_lint_auto extensions should abide to ament_lint_auto conventions in doing so. Namely, linter hooks should abide to the CMake list variable `AMENT_LINT_AUTO_FILE_EXCLUDE` which will propagate a file exclusion list to each linter invoked by ament_lint_auto.

Closes #343.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/ament_cmake_copyright/cmake/ament_cmake_copyright_lint_hook.cmake
+++ b/ament_cmake_copyright/cmake/ament_cmake_copyright_lint_hook.cmake
@@ -15,4 +15,9 @@
 message(
   STATUS
   "Added test 'copyright' to check source files copyright and LICENSE")
-ament_copyright()
+
+if(DEFINED AMENT_LINT_AUTO_FILE_EXCLUDE)
+  ament_copyright(EXCLUDE ${AMENT_LINT_AUTO_FILE_EXCLUDE})
+else()
+  ament_copyright()
+endif()
--- a/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
+++ b/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake
@@ -46,6 +46,10 @@ if(_source_files)
     list(APPEND _all_exclude ${ament_cmake_cppcheck_ADDITIONAL_EXCLUDE})
   endif()
 
+  if(DEFINED AMENT_LINT_AUTO_FILE_EXCLUDE)
+    list(APPEND _all_exclude ${AMENT_LINT_AUTO_FILE_EXCLUDE})
+  endif()
+
   # BUILDSYSTEM_TARGETS only supported in CMake >= 3.7
   if(NOT CMAKE_VERSION VERSION_LESS "3.7.0")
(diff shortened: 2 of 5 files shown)


aprotyas:
@methylDragon pinging for a review!

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16836/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11414/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17243/)

methylDragon (approved):

clalancette:
@Mergifyio backport humble

PR #403, opened 2022-08-22
Enable the user of a misra.json config file if the user provides one

mjeronimo:
The MISRA error text is copyrighted. However, if the user puts together a misra.json file that points to a file that contains the MISRA rules text and puts these in ament_cppcheck's package share directory, the utility will pick up the rules text file and output the correct error messages rather than placeholders that only specify the MISRA rule number. 

### DIFF
--- a/ament_cppcheck/ament_cppcheck/main.py
+++ b/ament_cppcheck/ament_cppcheck/main.py
@@ -167,17 +167,19 @@ def main(argv=sys.argv[1:]):
     if jobs:
         cmd.extend(['-j', '%d' % jobs])
     if args.enable_misra_checks:
-        # The MISRA rule text is copyrighted. If the user happens to provide a misra.json configuration
-        # file that specifies a rule text file to use, use it. Here's an example misra.json:
+        # The MISRA rule text is copyrighted. However, if the user happens to have a misra.json
+        # configuration file in the $HOME directory, use it.
+        #
+        # Here's an example misra.json:
         # {
         #   "script": "/usr/lib/x86_64-linux-gnu/cppcheck/addons/misra.py",
         #   "args": [
-        #      "--rule-texts=/home/spaceros-user/src/spaceros/install/ament_cppcheck/share/ament_cppcheck/misra-rules.txt",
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> puts these in ament_cppcheck's package share directory

Do the json and rules files have to be in the share directory of ament_cppcheck? In general I don't think it makes sense to encourage manually creating files in a package's FHS directories. In fact the only situation that comes to mind when it even makes sense for packages *other* than the target package to update a package-specific FHS subdirectory is for a package which extends the base package. When using ament_cppcheck from the apt repositories, placing files in, for example, `/opt/ros/rolling/share/ament_cppcheck` is also something only the root user could do.

Additionally, requiring these files be present there means that each workspace would need its own misra.json and rules text whereas allowing the user to specify a file could mean they keep one copy of each in their home directory.

mjeronimo:
@nuclearsandwich Updated to look in the user's HOME directory instead.

ament_cppcheck/ament_cppcheck/main.py
```diff
@@ -166,7 +167,20 @@ def main(argv=sys.argv[1:]):
...
+        # Here's an example misra.json:
+        # {
+        #   "script": "/usr/lib/x86_64-linux-gnu/cppcheck/addons/misra.py",
+        #   "args": [
+        #      "--rule-texts=/home/spaceros-user/misra-rules.txt",
+        #      "--suppress-rules 17.3,21.12"
+        #   ]
+        # }
+
+        misra_json = os.path.join(os.environ['HOME'], 'misra.json')
```
nuclearsandwich, line 182:
```suggestion
        misra_json = os.path.join(os.path.expanduser('~'), 'misra.json')
```

Thanks for iterating. Using `os.path.expanduser` will work on Windows as well and fall back to looking up the current user in `/etc/passwd` if `HOME` is unset for some reason. https://docs.python.org/3/library/os.path.html#os.path.expanduser

Accepting this suggestion will also require adding `import os.path` to the imports at the top of the file.

nuclearsandwich (approved):
Approved with one suggestion for better cross-platform support.

Issue #415, opened 2022-10-17
Restrict allowed licenses in `ament_copyright`


alsora:
Hi, would it be possible to add the following features to `ament_copyright` ?
 - Restrict the set of known licenses to a subset of the existing ones.

Currently this package defines a bunch of valid licenses.
I would like to run a check that requires only 1 specific license to be present in my project.

Issue #414, opened 2022-10-17
`ament_copyright --add-copyright-year` broken


alsora:
The `ament_copyright --add-copyright-year` tool is currently broken.

 - https://github.com/ament/ament_lint/blob/rolling/ament_copyright/ament_copyright/main.py#L277
 this attribute should be named `copyright_identifiers` (with an `s` at the end)
 - https://github.com/ament/ament_lint/blob/rolling/ament_copyright/ament_copyright/main.py#L290
 the function `search_copyright_information` returns 2 values, not 3

Besides these obvious bugs, I'm also unsure that the `add_copyright_year` function should parse the copyright comments.
the parsed data should already be available in the descriptor objects.

fixing these bugs should be easy.
however, it looks like the tool supports multiple copyrights. How do we deal with them?

Let's say that I have a file with 2 copyrights

```
# Copyright 2021 TheNewCompany

# Copyright 1999 TheOldCompany
```

IMO we should have the tool update only the latest copyright (or maybe the first one found in the document)

PR #416, opened 2022-10-17
various fixes to ament-copyright and add --allowed-licenses argument

alsora:
Signed-off-by: Alberto Soragna <alberto.soragna@gmail.com>

This PR addresses the bugs mentioned in https://github.com/ament/ament_lint/issues/414.
This PR adds the feature described in https://github.com/ament/ament_lint/issues/415

### DIFF
--- a/ament_copyright/ament_copyright/main.py
+++ b/ament_copyright/ament_copyright/main.py
@@ -29,11 +29,11 @@ from ament_copyright import SOURCE_FILETYPE
 from ament_copyright import UNKNOWN_IDENTIFIER
 from ament_copyright.crawler import get_files
 from ament_copyright.parser import get_comment_block
+from ament_copyright.parser import get_copyright_information_regex
 from ament_copyright.parser import get_index_of_next_line
 from ament_copyright.parser import parse_file
 from ament_copyright.parser import scan_past_coding_and_shebang_lines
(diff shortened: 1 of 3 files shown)


superdyoll:
What's stopping this branch from being merged? Just re-discovered this bug myself and it'd be nice if this bug was fixed. It's hidden in the released version of ament_copyright due to the typo in add_copyright_year in main.py of the missing s in copyright_identifiers that this merge request fixes. Would prefer that ament_copyright updates the copyright year itself rather than having to write a separate script to do it.

clalancette (changes requested):
I think this PR does a little too much, in my opinion.  To get this in, I think 3 PRs would help:

1.  Adds in `--allowed-licenses`, along with the changes to support that feature.
2.  A bugfix for the "copyright_identifiers" bug (this will be easy to backport)
3.  The fixes to change to a regex for the copyright information

Split up that way, I think we could get this in.

superdyoll:
@clalancette have raised a new PR here https://github.com/ament/ament_lint/pull/466 to fix the bug in --add-copyright-year caused by the lack of s on identifiers. Unfortunately it wasn't as simple as just typing s as the code had to be updated to work with multiple copyright notices at the same time.

PR #419, opened 2022-11-01
Fixed a few issues with parsing and SARIF gen for SpaceROS

Ronoman:
1. Before, the `rules` field in the SARIF output was polluted with many duplicate rules, as the full error message was included. Now, only the rule id that `clang-tidy` outputs between brackets (i.e. `[google-explicit-constructor]`) is taken and stored in rules.
2. Before, artifact paths sometimes had part or all of the error message in them. No more!
3. Before, the `startLine` and `startColumn` fields were strings. These are now integers.
4. Before, the result message had a lot more information than was needed. Now, it is just the error message (no location data or rule id).

This was tested on `rclcpp` and `rcutils`. `ament_clang_tidy` ran successfully on both, and outputted valid SARIF. However, I did have to change some of the RegEx's that were parsing `clang_tidy` output. I don't have a great way to check if I'm truly capturing all of the valid output of `clang_tidy`, so this may be unintentionally hiding some violations.


ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -214,8 +214,14 @@ def start_subprocess(full_cmd):
     for filename in files:
         report[filename] = []
 
-    error_re = re.compile('(/.*?\\.(?:%s)):(\\d+):(\\d+): (?:warning:|error:)' %
-                          '|'.join(extensions))
+    # Regex explanation:
+    #   \s*\^?                   : Capture many spaces, then an optional `^`. clang_tidy output isn't great.
```
nuclearsandwich, line 218: `*` is used to capture 0 or more spaces. Is 0 valid or should it be a `+` to capture 1 or more?

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -214,8 +214,14 @@ def start_subprocess(full_cmd):
...
+    #   (\/home\/^[^:]*)         : Group capture. Everything from `/home/` up until a `:`.
```
nuclearsandwich, line 219:
```suggestion
    #   (/home/^[^:]*)         : Group capture. Everything from `/home/` up until a `:`.
```

I don't think that `/` is a metacharacter outside of tools like `sed` when it is used as a pattern delimiter.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -214,8 +214,14 @@ def start_subprocess(full_cmd):
...
+    #   (\d+)                    : Group capture. Grabs line number.
+    #   (\d+)                    : Group capture. Grabs column number.
+    #   (?:warning:|error:note:) : Non-capturing group. Matches warning, error, note.
```
nuclearsandwich, line 222:
```suggestion
    #   (?:warning:|error:|note:) : Non-capturing group. Matches warning, error, note.
```

Typo in the comment.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -214,8 +214,14 @@ def start_subprocess(full_cmd):
...
+    #   \[(.*)\]                 : Matches and captures [<rule_name>]. Ignores any messages from clang_tidy without an ending [<rule_name>].
+    error_re = re.compile('\\s*\^?\/home\/([^:]*):(\\d+):(\\d+): (?:warning:|error:|note:).*\\[(.*)\\]')
```
nuclearsandwich, line 224: Nice inclusion of the regex segments explanation 🙇🏼

```suggestion
    # Regex explanation:
    #   \s*\^?                   : Capture many spaces, then an optional `^`. clang_tidy output isn't great.
    #   (\/home\/^[^:]*)         : Group capture. Everything from `/home/` up until a `:`.
    #   (\d+)                    : Group capture. Grabs line number.
    #   (\d+)                    : Group capture. Grabs column number.
    #   (?:warning:|error:note:) : Non-capturing group. Matches warning, error, note.
    #   \[(.*)\]                 : Matches and captures [<rule_name>]. Ignores any messages from clang_tidy without an ending [<rule_name>].
    error_re = re.compile(r'\s*\^?/home/([^:]*):(\d+):(\d+): (?:warning:|error:|note:).*\[(.*)\]')
```

I think the regex pattern is easier to read when using a raw string. I also removed a set of spurious escapes of `/` which is not a metacharacter as mentioned above.

ament_clang_tidy/ament_clang_tidy/main.py
```diff
@@ -443,20 +449,22 @@ def get_sarif_content(report, clang_tidy_version):
...
+            new_rule = {
                 'id': rule_id,
-                'shortDescription': {
-                    'text': description.rstrip()
-                },
                 'helpUri': 'https://clang.llvm.org/extra/clang-tidy/checks/list.html',
-            })
+            }
+            if new_rule not in rules:
+                rules.append(new_rule)
```
nuclearsandwich, line 457: How much overlap does this operation have with the deduplication post-processing in process_sarif?

I would prefer to keep the ament output as "raw" as possible from the wrapped tool.

nuclearsandwich (commented):
Initial review is positive! A few comments and observations.

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: I _think_ `*` is what I want here. This was added to fix output from `clang_tidy` that looks like this:

```
/usr/lib/llvm-10/lib/clang/10.0.0/include/stdarg.h:27:29: note: expanded from macro 'va_copy'
#define va_copy(dest, src)  __builtin_va_copy(dest, src)
                            ^/home/spaceros-user/src/spaceros/src/rcutils/src/process.c:98:3: warning: Call to function 'memcpy' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'memcpy_s' in case of C11 [clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling]
  memcpy(executable_name, bname, baselen);
  ^
  ```
  
  Notice that on the third line, there are many spaces preceding a `^`, then immediately a path for a completely different static analysis result. There are many cases where there are no spaces and no `^`, so `\s*\^?` is what we want. A more explicit way to write it might be `(?\s*\^)?`, as a non-capturing group that is optional.

File: ament_clang_tidy/ament_clang_tidy/main.py (hunk 5 above)
Comment (Ronoman) [reply ament_clang_tidy/ament_clang_tidy/main.py:457]: The duplication that `process_sarif` handles is within the `results` field. This is preventing duplication on the rules types reported by `clang_tidy`, which ends up in the `runs[0].tool.rules` field of the SARIF. I could see an argument for porting this over to `process_sarif`, but I don't believe this is duplicating work.

File: ament_clang_tidy/ament_clang_tidy/main.py (hunk 1 above)
Comment (nuclearsandwich) [reply ament_clang_tidy/ament_clang_tidy/main.py:218]: Yeah, thanks for including the example output, that helped a lot. Since that caret indicates the column in the above line, it's possible for that to appear at column 0 and lack any preceding spaces so the `*` 0 or more metacharacter is the correct match.

File: ament_clang_tidy/ament_clang_tidy/main.py [hunk 6]
```diff
@@ -214,8 +214,14 @@ def start_subprocess(full_cmd):
     for filename in files:
         report[filename] = []
 
-    error_re = re.compile('(/.*?\\.(?:%s)):(\\d+):(\\d+): (?:warning:|error:)' %
-                          '|'.join(extensions))
```
nuclearsandwich, line 218: The inclusion of extensions in the compiled pattern appears to've been dropped with these changes. Is that intentional?

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: https://regexr.com tricked me into thinking the extra backslashes for `/home/` were necessary, but I think they can be dropped.

Ronoman:
It seems like I made a poor assumption with some of the output, but I'm not sure how to resolve it. Most `clang_tidy` errors look like this:

```
/home/spaceros-user/src/spaceros/src/rcutils/src/char_array.c:153:14: warning: Call to function 'vsnprintf' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'vsnprintf_s' in case of C11 [clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling]
  int size = vsnprintf(char_array->buffer, char_array->buffer_capacity, format, args_clone);
             ^
```

These are matched and groups are extracted properly with the regex as it stands. However, some output looks like this:

```
/home/spaceros-user/src/spaceros/src/rcutils/src/char_array.c:123:5: note: Call to function 'memcpy' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'memcpy_s' in case of C11
```

Here, the rule ID is missing from the result. In most cases, these are preceded by the same exact error, like this block:

```
/home/spaceros-user/src/spaceros/src/rcutils/src/array_list.c:175:5: warning: Call to function 'memcpy' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'memcpy_s' in case of C11 [clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling]
    memcpy(dst_ptr, src_ptr, array_list->impl->data_size * copy_count);
    ^
/home/spaceros-user/src/spaceros/src/rcutils/src/array_list.c:175:5: note: Call to function 'memcpy' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'memcpy_s' in case of C11
```

In these cases, it's safe to throw away the second result since it is duplicating one already listed (note that both results are pointing to `array_list.c:175:5`). However, this isn't always the case. How can we capture this "incorrect" `clang_tidy` output properly in SARIF?

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: Good catch! Will drop this in a commit once I test to make sure functionality is the same.

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: It is intentional. The pattern before was matching a filename and then an extension before moving to the `:<row>:<col>` sections of the output. The new regex will match paths that start with `/home/`, then any number of characters that is not a `:` which indicates the start of the `:<row>:<col>` block. I believe this matches the same strings as before, and is slightly less complex.

ament_clang_tidy/ament_clang_tidy/main.py
mjeronimo: I don't think the rules should be repeated in the SARIF file. Multiple discovered issues can reference the same rule.

mjeronimo (approved):

ament_clang_tidy/ament_clang_tidy/main.py
nuclearsandwich: > The new regex will match paths that start with /home/, then any number of characters that is not a `:`...

You know, I saw this in the regexp and didn't question it, and it is already in the code but.. does this mean that errors outside of the /home/ path will not be captured?

Some of our builds occur in `/root/` or `/src/`.

nuclearsandwich:
> In these cases, it's safe to throw away the second result since it is duplicating one already listed (note that both results are pointing to `array_list.c:175:5`). However, this isn't always the case. It seems like `clang_tidy` does not add the rule ID when it is reporting `note:` results, as opposed to `warning:` or `error:` results. How can we capture this different `clang_tidy` output properly in SARIF?

It doesn't look like clang-tidy itself supports formatting its output nor does its list of checks state what level a check is is (note, warning, or error). The ruleID is optional right? Can we just omit it for notes?

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: It does mean those won't be captured. What are the cases in which builds happen under `/root` or `/src`? Happy to swap this regex to be more accommodating. Should we match `(?:/home|/root|/src)`, or should it be more permissive and require a preceding `/` but not make any other assumptions about the path?

Ronoman:
> It doesn't look like clang-tidy itself supports formatting its output nor does its list of checks state what level a check is is (note, warning, or error). The ruleID is optional right? Can we just omit it for notes?

According to [The Spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541083), it looks like we can omit `result.ruleId` in some cases. The relevant sentence (I think) is this:

`If theDescriptor does not exist (that is, if theTool does not contain a reportingDescriptor object (§3.49) that describes the rule that was violated), then rule SHALL NOT be present.`

`theTool` _probably_ does have the relevant rule ID, but since the `result` for `note`s doesn't give it to us, I think we're okay to omit it.

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: Resolved in 5d5e2eda

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: Resolved in 5d5e2eda

ament_clang_tidy/ament_clang_tidy/main.py
Ronoman: Resolved in 5d5e2eda

Ronoman:
Resolved all outstanding comments in 5d5e2eda. However, a new slight issue arises...

I've updated the regex rules to capture all clang_tidy results, even one that don't end in a `[rule description]`. However, this means that `process_sarif` is no longer going to de-duplicate results in clang_tidy properly. on [line 106](https://github.com/space-ros/process_sarif/blob/main/process_sarif/sarif_helpers.py#L106) of sarif_helpers, we check if the tuple (threeple?) `(ruleId, artifact, region)` has been seen before. With these changes, there are now results in clang_tidy that are exact duplicates, but some are missing the ruleId, so this check is no longer sufficient.

Is it okay to reduce the tuple to `(artifact, region)`? This would mean that any results that are pointing to the same file, with the exact same region (line number, sometimes column number) will be considered identical. I haven't yet observed two unique results pointing to the exact same artifact and region, so this seems like a reasonable assumption, but not always guaranteed to be true.

mjeronimo:
Perhaps it would be more accurate to keep the (ruleId, artifact, region) and only make the reduction to (artifact, region) if necessary. In other words, ruleId would be optional in the comparision function, but used if present in both items to be compared.

ivanperez-keera:
I'm following up on this (we have a PR in Space ROS that depends on this).

Does it still make sense to do this? If so, what's missing?

EzraBrooks:
clang-tidy can be configured to output a YAML-formatted "fixes file" which is probably the way I'd recommend to ingest tidy output programmatically.

mjcarroll:
Closing as stale. Feel free to re-open.

PR #420, opened 2022-11-04
Declare dependencies on expected flake8 plugins.

nuclearsandwich:
These flake8 plugins have been in the from-source setup instructions for ROS 2 a long time. For Ubuntu 22.04 we finally got all of them packaged so we can start removing pip packages from our installation requirements.

I think this change should be back-portable to Humble but it can't go back to Focal because not all of these packages are available there.

This PR requires https://github.com/ros/rosdistro/pull/35197

### DIFF
--- a/ament_flake8/package.xml
+++ b/ament_flake8/package.xml
@@ -19,6 +19,14 @@
 
   <exec_depend>ament_lint</exec_depend>
   <exec_depend>python3-flake8</exec_depend>
+  <exec_depend>python3-flake8-blind-except</exec_depend>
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
@ros-pull-request-builder retest this please

vonZeppelin:
Nice, it'll simplify CI setups! Would be great to have as part of the upcoming release – any reasons to delay the merge?

clalancette:
So we partially did this in #454, so closing this out as a duplicate.

PR #426, opened 2022-11-22
ament_clang_format: use open braces for enum definitions

james-rms:
The ROS 2 C++ style guide is [ambiguous about this](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#open-versus-cuddled-braces):

> - Use open braces for function, class, and struct definitions, but cuddle braces on if, else, while, for, etc… 
Exception: when an if (or while, etc.) condition is long enough to require line-wrapping, then use an open brace (i.e., don’t cuddle).

It seems more consistent to also use open braces for enum declarations, ie:
```cpp
enum struct Format
{
  OPTION_ONE,
  OPTION_TWO,
};
```

### DIFF
--- a/ament_clang_format/ament_clang_format/configuration/.clang-format
+++ b/ament_clang_format/ament_clang_format/configuration/.clang-format
@@ -9,6 +9,7 @@ BraceWrapping:
   AfterFunction: true
   AfterNamespace: true
   AfterStruct: true
+  AfterEnum: true
 BreakBeforeBraces: Custom
 ColumnLimit: 100
 ConstructorInitializerIndentWidth: 0


mjeronimo:
I'll bring this up next week with the ROS 2 team and get back to you. Perhaps we should clarify the style guide as well.

clalancette:
> I'll bring this up next week with the ROS 2 team and get back to you. Perhaps we should clarify the style guide as well.

I took a look through our code, and we are already using open-braces for enums.  It's enforced by our uncrustify rules.  We've also already updated the style guide in https://github.com/ros2/ros2_documentation/pull/3203 .

As far as I know, none of the ROS 2 core packages use `ament_clang_format`, so this shouldn't change anything there (though we'd want to do a full CI run just to be sure).  The bigger concern is downstream packages that use this; they very well may get new warnings from this.

At the end of the day, I'm OK with that as that is the point of Rolling.  But because of that, I don't think that this would be backportable to Humble or other releases.

So this gets a thumbs up from me with green CI, assuming there are no other technical problems.

mjeronimo:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17694/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12240/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18299/)

james-rms:
FYI the reason I made this PR in the first place is that without `AfterEnum: true`, clang-format on Foxy and Galactic uses open braces for enums but Humble and Rolling use cuddled. See this CI run for evidence: https://github.com/ros-tooling/rosbag2_storage_mcap/actions/runs/3519952298/jobs/5900411262

For that reason I think it makes sense to backport this to Humble - this makes clang-format's behavior across Foxy, Galactic, Humble and Rolling the same.

That said, I see the concerns about breaking people's existing formatting and don't feel strongly about this, so it's up to you.

PR #439, opened 2023-05-11
add ament_bandit wrapper

florcabral:
Contribute a new wrapper for the Bandit static analyzer.

### DIFF
--- a/ament_bandit/README.md
+++ b/ament_bandit/README.md
@@ -1,0 +1,54 @@
+# ament_bandit
+
+Analyzes ``.py`` files using the [Bandit](https://bandit.readthedocs.io/) static analyzer.
+
+## Install
+
+`cd ament_bandit`
+
+`python3 setup.py install`
+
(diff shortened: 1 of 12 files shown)


clalancette:
My major question with this PR is: does this need to be in the `ament_lint` core?  That is, it would be much easier to just release this into ROS 2 as a separate package, which would:

1. Make it easier to make changes (you wouldn't have to wait for core maintainers to have time to review things).
2. Not subject it to the stricter rules of being in the core.
3. Reduce the support burden on the ament_lint maintainers.

Thoughts?

tfoote:
I think it would be good to release this as a separate package initially. That way it can be used, validated and iterated upon with at a faster rate. Once it's more mature we can consider bringing it into the default installation and into this repo too.

clalancette:
Given the previous comments, I'm going to close this out for now.  But do please consider releasing `ament_bandit` as a standalone package.

Issue #411, opened 2022-09-20
[ament_clang_tidy] Does not print errors and warnings in the terminal


AugusteBourgois:
Hi there,

I am working on Ubuntu 22.04, with ROS 2 humble. I have a workspace with a single package, for which I have generated a `compile_commands.json` file in the `build` folder.

I am running the following command : `ament_clang_tidy --config <path to my .clang-tidy file> build/my_package`.

The only output I get from this command is 
`The invocation of "clang-tidy" failed with error code 1: Command '['/usr/bin/clang-tidy', '-p', 'build/my_package', "--config={CheckOptions: [{key: readability-identifier-naming.NamespaceCase, value: lower_case}, {key: readability-identifier-naming.ClassCase, value: CamelCase}, {key: readability-identifier-naming.PrivateMemberPrefix, value: m_}, {key: readability-identifier-naming.StructCase, value: CamelCase}, {key: readability-identifier-naming.FunctionCase, value: lower_case}, {key: readability-identifier-naming.VariableCase, value: lower_case}, {key: readability-identifier-naming.GlobalConstantCase, value: UPPER_CASE}], Checks: '-*, readability-identifier-naming, readability-magic-numbers\n\n    ', WarningsAsErrors: '*'}\n", '--header-filter', 'include/my_package/.*', 'path to my file.cpp']' returned non-zero exit status 1.`

It seems to me that this is due to the way `subprocess.check_output` is used in the `main.py` file. Indeed, when the error code of the command is non-zero, it raises an exception which contains the desired output (the string describing the errors and warnings), but the output of `subprocess.check_output` does not contain anything.

A quick fix would be to add the following line below `except subprocess.CalledProcessError as e:`:
`output = e.output.decode()`

Simply adding this fixes my problem : the output is printed in the terminal as expected.

Kind regards

Issue #350, opened 2022-01-17
[ament_cmake_clang_tidy] failed test is detected as passed


wmmc88:
This is using the latest master of ament_lint and latest rolling release.

From https://github.com/uwrobotics/uwrt_mars_rover/runs/4834656922?check_suite_focus=true#step:4:1520 :
```
  test 2
      Start 2: clang_tidy
  
  2: Test command: /usr/bin/python3 "-u" "/opt/ros/galactic/share/ament_cmake_test/cmake/run_test.py" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/build/uwrt_mars_rover_drivetrain_hw/test_results/uwrt_mars_rover_drivetrain_hw/clang_tidy.xunit.xml" "--package-name" "uwrt_mars_rover_drivetrain_hw" "--output-file" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/build/uwrt_mars_rover_drivetrain_hw/ament_clang_tidy/clang_tidy.txt" "--command" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/install/ament_clang_tidy/bin/ament_clang_tidy" "--xunit-file" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/build/uwrt_mars_rover_drivetrain_hw/test_results/uwrt_mars_rover_drivetrain_hw/clang_tidy.xunit.xml" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/build/uwrt_mars_rover_drivetrain_hw" "--config" "/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/src/tsv4w6qdap/uwrt_mars_rover/uwrt_mars_rover_drivetrain/uwrt_mars_rover_drivetrain_hw/../../.clang-tidy"
  2: Test timeout computed to be: 300
  2: -- run_test.py: invoking following command in '/__w/uwrt_mars_rover/uwrt_mars_rover/ros_ws/src/tsv4w6qdap/uwrt_mars_rover/uwrt_mars_rover_drivetrain/uwrt_mars_rover_drivetrain_hw':
[...]

PR #452, opened 2023-09-08
Clang tidy add extra arg option and return 1 when any error is catched.

wuisky:
Hi there. Thank you for developing such nice tool.
However, i noticed that rclcpp use some c++17 features, 
and without passing --extra-arg=-std=c++17 to clang-tidy it will report errors in these headers.
After this patch, we can pass extra-arg just like:
```bash
$ ament_clang_tidy path_to_compile_commands.json  --extra-arg=-std=c++17 
```
Also, i fix issues mentioned in #350 #411.
Hope this helps.


pucciland95:
A gentle ping since this is a particularly important update to let `ament_clang_tidy` be usable. Could it also be ported to `Humble` ?

Cheers,

Niccolo

wuisky:
@clalancette could you take a look of this PR?

acarrillo-civ:
This would be a great improvement, making `ament_clang_tidy` significantly easier to use if you want your build to fail on lint errors.

adivardi:
Is there any update on that? Without that, running ament_clang_tidy in CI is not useful, since it will always pass

PR #456, opened 2023-10-06
(ament_xmllint) add extensions argument

### LINKED ISSUE #455: (ament_xmllint) add extension argument
Allow for the extensions to be set. So xacro files etc can also be checked.

MatthijsBurgh:
Fixes #455

### DIFF
--- a/ament_xmllint/ament_xmllint/main.py
+++ b/ament_xmllint/ament_xmllint/main.py
@@ -29,7 +29,7 @@ from xml.sax.saxutils import quoteattr
 
 
 def main(argv=sys.argv[1:]):
-    extensions = ['xml']
+    default_extensions = ['xml']
 
     parser = argparse.ArgumentParser(
         description='Check XML markup using xmllint.',
@@ -39,8 +39,9 @@ def main(argv=sys.argv[1:]):
         nargs='*',
         default=[os.curdir],
         help='The files or directories to check. For directories files ending '
-             'in %s will be considered.' %
-             ', '.join(["'.%s'" % e for e in extensions]))
+             'in %s will be considered. Unless overruled by the --extensions '
+             'argument' %
+             ', '.join(["'.%s'" % e for e in default_extensions]))
     parser.add_argument(
         '--exclude',
         nargs='*',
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
friendly ping @mjeronimo

Or maybe @clalancette @rodericktaylor?

ament_xmllint/ament_xmllint/main.py
```diff
@@ -39,15 +39,21 @@ def main(argv=sys.argv[1:]):
         nargs='*',
         default=[os.curdir],
         help='The files or directories to check. For directories files ending '
```
clalancette, line 41:
```suggestion
        help='The files or directories to check. For directories, only files ending '
```

(I know you didn't add this, but I think we can improve the wording a bit while we are here)

ament_xmllint/ament_xmllint/main.py
```diff
@@ -39,15 +39,21 @@ def main(argv=sys.argv[1:]):
...
-             'in %s will be considered.' %
-             ', '.join(["'.%s'" % e for e in extensions]))
+             'in %s will be considered. Unless overruled by the --extensions '
+             'argument' %
```
clalancette, line 43:
```suggestion
             'in %s will be considered (unless overruled by the --extensions '
             'option)' %
```

ament_xmllint/ament_xmllint/main.py
```diff
@@ -39,15 +39,21 @@ def main(argv=sys.argv[1:]):
...
         '--exclude',
         nargs='*',
         default=[],
         help='Exclude specific file names and directory names from the check')
     # not using a file handle directly
     # in order to prevent leaving an empty file when something fails early
+    parser.add_argument(
+        '--extensions',
+        nargs='*',
+        default=default_extensions,
```
clalancette, line 55:
```suggestion
        default=['xml'],
```

(and then we can remove the line of code at the beginning of this function)

clalancette (changes requested):
Overall, I think this is a fine change to have.  I've left some feedback for improvement.

ament_xmllint/ament_xmllint/main.py
MatthijsBurgh: No, as `default_extensions` is also used to generate the help of the `paths` option.

I don't think we want to maintain that list in two spots. Or we remove the extensions from that help msg.

clalancette (approved):
Looks reasonable to me, thanks.  I'll run CI on it next.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19676/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14183/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20400/)

PR #459, opened 2023-11-08
Changes to make ament_flake8 work with v6+

mjcarroll:
`parse_preliminary_options_and_args` and `parse_preliminary_options` have both been removed in favor of `parse_args` in this commit: https://github.com/PyCQA/flake8/commit/0d667a73299971f1cf8ff549c519fffb282b1faf

(No corresponding PR as far as I can tell).

This is the minimal set of changes to get build/test passing on flake8 v6

### DIFF
--- a/ament_flake8/ament_flake8/main.py
+++ b/ament_flake8/ament_flake8/main.py
@@ -143,6 +143,7 @@ def get_flake8_style_guide(argv):
     # appropriate options to pass into the standard flake8.legacy.get_style_guide();
     # passing argv gets it to determine the options for us.
     application = flake8_app.Application()
+
     if hasattr(application, 'parse_preliminary_options'):
         prelim_opts, remaining_args = application.parse_preliminary_options(
             argv)
@@ -170,7 +171,7 @@ def get_flake8_style_guide(argv):
             )
             application.register_plugin_options()
             application.parse_configuration_and_cli(cfg, cfg_dir, remaining_args)
-    else:
+    elif hasattr(application, 'parse_preliminary_options_and_args'):
         application.parse_preliminary_options_and_args([])
         flake8.configure_logging(
             application.prelim_opts.verbose, application.prelim_opts.output_file)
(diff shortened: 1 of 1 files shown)


abrzozowski:
Hi @mjcarroll, I hope the dependency of `ament_flake8` on `flake8` will be updated soon.
What do you think about explicitly defining which version of `flake8` is supported by `ament_flake8` in setup.py ([ament_flake8/setup.py](https://github.com/ament/ament_lint/blob/master/ament_flake8/setup.py))?

ament_flake8/ament_flake8/main.py
```diff
@@ -199,7 +210,8 @@ def parse_config_file(config_file):
         opts_manager = manager.OptionManager(
             version=flake8_version,
             plugin_versions='',
-            parents=[]
+            parents=[],
+            formatter_names=[],
```
clalancette, line 214: I don't think that this works for version 5 of flake8: https://github.com/PyCQA/flake8/blob/5.0.4/src/flake8/options/manager.py#L318-L323

So I think we need another conditional here which adds this for version 6, but not for version 5.

clalancette (changes requested):

ament_flake8/ament_flake8/main.py
tylerjw: I tried building and testing rolling with this change with flake8 v5 and then separately v6 in my python environment and it seems to work. Is flake8 vendored in ros? Is there some way to know which flake8 it is using?

In the case of v6 and v5 I see this warning:

```
test/test_flake8.py::test_flake8
  /usr/lib/python3/dist-packages/flake8/plugins/manager.py:261: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    eps = importlib_metadata.entry_points().get(self.namespace, ())

-- Docs: https://docs.pytest.org/en/stable/warnings.html
---
```

clalancette: I just tested this out locally, and indeed this did not work with flake8 5 anymore.  I've now pushed fdb6977984986fb94e3e083095e997ee5c010770, which fixes that problem.

clalancette (approved):

clalancette:
@Mergifyio backport iron humble

clalancette:
@Mergifyio backport iron

PR #474, opened 2024-02-07
Update maintainer list in package.xml files

mjeronimo:
I'm no longer a maintainer and don't need to be credited as an author.


christophebedard:
I think this should be changed here: https://github.com/audrow/update-ros2-repos/tree/main/apps/update-maintainers/data. For example, see #421.

Otherwise, the `CODEOWNERS` file should also be updated.

clalancette:
Given that this is just an update to the package.xml metadata, I'm going to rely on the Rpr job and merge this in.

PR #476, opened 2024-02-22
Add config file support to ament_pep257

### LINKED ISSUE #413: ament_pep257 support for user defined config file
I want to adjust some of the pep257 behavior in our project. In ros2 there seem to be some ignored styles configured here: https://github.com/ament/ament_lint/blob/rolling/ament_pep257/ament_pep257/main.py#L43

From the official pep257/pydocstyle documentaton, there should be an option to use an additional configuration file: https://pep257.readthedocs.io/en/1.0.0/usage.html#configuration-files

I tried placing these in my ros workspace but they do not seem to be recoginized. Is there a way to do this?

InvincibleRMC:
I was trying to use a config file with ament_pep257 but, realized there was no implementation. Fixes #413.

### DIFF
--- a/ament_lint_auto/doc/index.rst
+++ b/ament_lint_auto/doc/index.rst
@@ -44,6 +44,26 @@ The documentation of the package `ament_cmake_test
 <https://github.com/ament/ament_cmake>`_ provides more information on testing
 in CMake ament packages.
 
+
+How to exclude linter modules with ament_lint_auto?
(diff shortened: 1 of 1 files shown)


InvincibleRMC:
Hello there maintainers. Is there something else I should? If you all are busy that is fine just wanted to make sure I am not missing anything since this is my first pull request to this repo.

PR #478, opened 2024-02-29
Try to find an flake8 config file in the local package

ahcorde:
It will be great to be able to define my own ament_flake8.ini configuration file locally, instead of using the default one.

This PR propose a fix. When a ROS Python package is created there is a folder called test with a `test_flake8.py` file that launches the test. I propose to put there your own config (if you want)
 - test/configuration/ament_flake8.ini

Otherwise I need to modify the `test_flake8.py` I think this solution is cleaner.

Thoughts?


ahcorde:
@audrow or @clalancette do you mind to review this one ?

PR #482, opened 2024-03-13
Add AMENT_LINT_AUTO_SKIP_PREEXISTING_TESTS flag

xfiderek:
## Summary
Added a flag `AMENT_LINT_AUTO_SKIP_PREEXISTING_TESTS` to every test in the repo. If the flag is defined and set to `ON`,  then already added tests (matched by name) are ignored.  This is contrary to default behavior of ament_lint_auto, as by default duplicate test names cause build failure.
The code is backward compatible -> added piece of code has no effect, unless the flag is set.

## Use case
This flag helps to solve https://github.com/space-ros/space-ros/issues/57.  The main idea for resolution of that issue is to inject additional cmake script during build using [CMAKE_PROJECT_INCLUDE](https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_INCLUDE.html) variable, that will add linters to packages. Check linked issue and [README in the PR](https://github.com/space-ros/space-ros/pull/145/files#diff-af72128dd2406e40947d93930b80ba6811d8bbbbbf338810733a77640b3be038) for more details.

### DIFF
--- a/ament_cmake_clang_format/cmake/ament_clang_format.cmake
+++ b/ament_cmake_clang_format/cmake/ament_clang_format.cmake
@@ -39,6 +39,15 @@ function(ament_clang_format)
     set(ARG_TESTNAME "clang_format")
   endif()
 
+  # https://cmake.org/cmake/help/latest/prop_dir/TESTS.html
+  get_directory_property(_declared_tests TESTS)
+  if(DEFINED AMENT_LINT_AUTO_SKIP_PREEXISTING_TESTS)
+    if((AMENT_LINT_AUTO_SKIP_PREEXISTING_TESTS) AND (${ARG_TESTNAME} IN_LIST _declared_tests))
+      message(VERBOSE "skipping test '${ARG_TESTNAME}' as it has already been added")
+      return()
+    endif()
+  endif()
+
   find_program(ament_clang_format_BIN NAMES "ament_clang_format")
   if(NOT ament_clang_format_BIN)
     message(FATAL_ERROR "ament_clang_format() variable 'ament_clang_format_BIN' must not be empty")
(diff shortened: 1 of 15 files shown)


fujitatomoya:
CC: @mjcarroll /assign

mjcarroll:
I'm not sure I fully understand what you are trying to accomplish here.  Can you expand a bit more on the use case and what you are trying to solve with this flag?

xfiderek:
Sure!

So basically we want to add functionality to spaceros, which will allow users to overwrite linters defined in CMakeLists.txt files of upstream packages, to ensure that all ros packages are analysed with the same set of linters. 

One way to solve that would be to modify these files directly. What we do instead is we inject cmake script during build process of upstream packages using CMAKE_PROJECT_INCLUDE variable. In that script, we define linters by using ament_lint_auto and calling find_package() on linters that we want to add (e.g. ament_cobra, ament_cppcheck).

Lets now imagine that an upstream package already includes one of our linters, lets say ament_cppcheck. Prior to this PR, the build would fail because the same test would have been defined twice -> once in our script and once in original CMakeLists.txt.
After this PR, when we set the flag to true, the test added in CMakeLists.txt will be ignored, hence avoiding duplication error.

Hope it makes sense. Ive tried to find another approach that will allow us to achieve this functionality, but i cant really come up with simpler and cleaner altenrative that would not involve modifying cmakelists.txt files of upstream packages.

mjcarroll:
Thanks.  I'm clear on the use case now.  Can you add an entry to the docs to describe the use of that variable.  What you have stated here is sufficient for me, just to make sure it is documented. 

I would recommend here: https://github.com/ament/ament_lint/blob/rolling/ament_lint_auto/doc/index.rst

xfiderek:
Hi, sorry for the delay on my side. Thanks a lot for the review, I have uploaded the documentation as you suggested.

xfiderek:
Hey @mjcarroll, could you take a look at my changes?

xfiderek:
@mjcarroll could you look into the updated docs so that we can merge this?

ivanperez-keera:
@mjcarroll I'm happy to merge the changes. I'd appreciate a final look just to make sure we are not missing anything here.

mjcarroll (approved):

PR #491, opened 2024-05-15
Fix error path for search_copyright_information.

clalancette:
Callers now expect it to return a 4-tuple, but in the case that there was no data it was only returning a 2-tuple.


ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21087/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15390/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21804/)

clalancette:
@Mergifyio backport jazzy

PR #494, opened 2024-05-29
clang-tidy: Add clang-tidy executable path argument

jjannik:
This MR allows an easy way to add clang-tidy caching
like here https://github.com/matus-chochlik/ctcache or https://github.com/ejfitzgerald/clang-tidy-cache/


clalancette (commented):
I guess my question here is: what is the use case?  Because it seems like we can already influence which `clang-tidy` binary is called by changing the PATH environment variable, so why do we need this?

jjannik:
True. As far as I can see you can not really influence the name of your executable. This would allow you to call something like `cache-clang-tidy` or whatever is expected for the above caching solutions. Which would make it clear to the caller that not actually clang tidy is used, more like a wrapper.

Issue #500, opened 2024-08-30
ament_cpplint Support for C++20


rodericktaylor:
ament_cpplint doesn't recognise c++20 headers and complains that C headers and complains:

`Found C system header after C++ system header. Should be: point_cloud_msg.h, c system, c++ system, other.`

On code like this.

```c++
#include <array>
#include <bit>
#include <cstdint>
#include <cstring>
#include <optional>
#include <ranges>
#include <span>
#include <stdexcept>
#include <string>
#include <string_view>
#include <utility>
```

PR #501, opened 2024-09-26
Retrofit a new version of cpplint.  Ignore Qt generated moc_ files.

rodericktaylor:
Addresses. #500


fujitatomoya:
@rodericktaylor thanks for the PR.

Can we divide this PR into 2 parts?

- Ignore checking moc generated cpp file by `ament_clang_tidy` package.
- Retrofit a new version of cpplint in `ament_cpplist` package. (against original issue https://github.com/ament/ament_lint/issues/500) especially this one, we need to see what happens to all downstream packages with new version by running CI before merging it.

rodericktaylor:
Yep. Will do.  I'll close this PR and create another.

PR #505, opened 2024-10-17
Force flake8 to the use the single-threaded pool.

clalancette:
The comment in the code has more information about why we want to do this.

This is a draft because a) I'm not 100% sure this fixes the issue, and b) I'm not sure what this does to our CI times.  @fujitatomoya FYI.

### DIFF
--- a/ament_flake8/ament_flake8/main.py
+++ b/ament_flake8/ament_flake8/main.py
@@ -264,6 +264,13 @@ def generate_flake8_report(config_file, paths, excludes, max_line_length=None):
     if max_line_length is not None:
         flake8_argv.append('--max-line-length={0}'.format(max_line_length))
 
+    # We've seen some problems, especially on RHEL-9, where using the multi-threaded
+    # pool in flake8 can cause the Python interpreter to crash.  Force flake8 to use
+    # the single-threaded pool here.  This has some performance implications for
+    # large codebases, but given the distributed nature of ROS packages this shouldn't
+    # generally be a large problem.
+    flake8_argv.append('-j=1')
+
     style = get_flake8_style_guide(flake8_argv)
 
     # Monkey patch formatter to collect all errors


fujitatomoya (commented):
👍 thanks, i will keep eyes on this!

* Linux-rhel Debug [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1535/)

fujitatomoya (approved):
i will give official lgtm because this turns everything into all green!!!

clalancette:
So it does indeed seem to make everything green, which is a good start.

In terms of job times, it looks like all jobs except one took the same, or shorter, than their nightly counterparts from last night.  The one exception was aarch64, which took 1 hr 34 min on the nightly, but 1 hr 55 min here.  I'm not sure why that is different.

That said, the jobs run here do not have a 1-1 correspondence with the nightly jobs.  The nightly jobs explicitly use either "Release" or "Debug", while the CI jobs use "None" (which is different than both of those).  So what I'm going to do here is to run another set of jobs, all in Release mode.  We'll see how that compares in terms of time.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22016/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16275/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22749/)

clalancette:
Hm, so the results on the Release jobs show that this is slower.  Note that in order to compare apples-to-apples, I actually took the total time for the job, subtracted off the time until we made it to the "Run Dockerfile" step, and used that as the numbers below (because the nightly jobs are at a disadvantage, since they *always* start from a fresh container):

| Job | nightly time | this PR time | delta |
|------|-------------------|-------------------|----------|
| Linux | 2 hr 33 min | 2 hr 34 min | +1 min |
| Linux aarch64 | 1 hr 29 min | 1 hr 57 min | +28 min |
| RHEL | 2 hr 14 min | 2 hr 18 min | +4 min |
| Windows | 4 hr 28 min | 4 hr 36 min | +8 min |

So it is clear that this change has something of an impact on CI times, particularly on aarch64.  However, aarch64 is the job that can most afford a CI time regression, since the workers aren't as heavily used as the amd64 workers, and it isn't ridiculously long like Windows is.

This one is a tough call to make.  My personal opinion is that we should take the CI time hit in favor of making RHEL consistently pass CI.  But pinging @cottsay @nuclearsandwich @Crola1702 @claraberendsen @ament/team for thoughts.

ament_flake8/ament_flake8/main.py
```diff
@@ -264,6 +264,13 @@ def generate_flake8_report(config_file, paths, excludes, max_line_length=None):
     if max_line_length is not None:
         flake8_argv.append('--max-line-length={0}'.format(max_line_length))
 
+    # We've seen some problems, especially on RHEL-9, where using the multi-threaded
+    # pool in flake8 can cause the Python interpreter to crash.  Force flake8 to use
+    # the single-threaded pool here.  This has some performance implications for
+    # large codebases, but given the distributed nature of ROS packages this shouldn't
+    # generally be a large problem.
+    flake8_argv.append('-j=1')
+
```
fujitatomoya, line 273: @clalancette just idea, can we do something like this? we have never seen this issue with other platform.

```suggestion
    import platform
    
    if platform.system() == "Linux" and "el9" in platform.version():
        # We've seen some problems, especially on RHEL-9, where using the multi-threaded
        # pool in flake8 can cause the Python interpreter to crash.  Force flake8 to use
        # the single-threaded pool here.  This has some performance implications for
        # large codebases, but given the distributed nature of ROS packages this shouldn't
        # generally be a large problem.
        flake8_argv.append('-j=1')
```

cottsay:
Have there been any attempts to force a different version of flake8 on RHEL 9? Or to force a Noble build to use the same flake8 as what RHEL 9 currently has?

It would be good to understand if this is a problem with the interpreter, some particular version(s) of flake8, a dependency thereof, or something else entirely.

Blast545:
Trying this PR again, as the problem in rhel started happening reliably again:

* Linux-rhel Debug [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1741/)

We closed https://github.com/ros2/ros2cli/issues/932 because the problem wasn't happening, but it started happening again. A quick check to changes in the repos / pip freeze, showed that the change triggering the problem could be either https://github.com/ros2/rosidl_dynamic_typesupport_fastrtps/pull/6 or a change in the pip version used from pip==21.2.3 to pip==21.3.1. cc: @Crola1702

Blast545:
> My personal opinion is that we should take the CI time hit in favor of making RHEL consistently pass CI

@clalancette maybe we don't have to take the time hit in other distributions? using fujitatomoya's suggestions on your PR, or something else, that only modifies rhel for this problem.

clalancette:
After more debugging, I found that this isn't necessary.  See https://github.com/ros2/ros2cli/pull/959 for a better workaround.

Issue #509, opened 2024-11-19
ament_mypy pytest is not working as intended and pass the CI when it shouldn't


tomkimsour:
I've tried to make colcon test fail for ament_mypy to confirm my doubt. 
In order to do so, I removed every typing annotation in the ament_mypy/ament_mypy/main.py
After, I on the cli `ament_mypy` and this error prompts : 
```
8 files checked
2 errors
ament_mypy/main.py: note: In function "_generate_mypy_report":
ament_mypy/main.py:141:5: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Success: no issues found in 8 source files


Checked files:

* setup.py
* ament_mypy/__init__.py
* ament_mypy/main.py
* ament_mypy/pytest_marker.py
* test/test_ament_mypy.py
* test/test_flake8.py
* test/test_mypy.py
* test/test_xmllint.py
```

There, we expect colcon tests to fail with the same behavior and raising 2 errors but if you run `colcon test --event-handlers console_direct+ --packages-select ament_mypy` the following output (please ignore flake8 errors) :

```
platform linux -- Python 3.10.12, pytest-8.0.2, pluggy-1.4.0
cachedir: /home/user/exchange/mono_repo/build/ament_mypy/.pytest_cache
rootdir: /home/user/exchange/mono_repo/src/ament_lint/ament_mypy
configfile: pytest.ini
plugins: ament-mypy-0.18.1, ament-xmllint-0.18.1, ament-copyright-0.18.1, ament-pep257-0.18.1, ament-flake8-0.18.1, ament-lint-0.18.1, launch-testing-1.0.6, launch-testing-ros-0.19.7, ament-black-0.2.4, anyio-4.4.0, repeat-0.9.3, rerunfailures-14.0, mock-3.6.1, cov-3.0.0, colcon-core-0.17.0
collecting ...
collected 12 items

test/test_ament_mypy.py .........                                        [ 75%]
test/test_flake8.py F                                                    [ 83%]
test/test_mypy.py .                                                      [ 91%]
test/test_xmllint.py .                                                   [100%]

```

tomkimsour:
After a quick glance at main.py of ament_mypy I noticed the  return value of the function is some kind of error code that is unrelated to the amount of error raised by the program.
If we want a similar behavior to ament_pep257 or ament_flake8. The return number of the main.py should be the amount of error found by the linter.

tomkimsour:
[...]

PR #510, opened 2024-11-19
Fix ament_mypy tests return code 

tomkimsour:
As mentionned in #509, main.py of ament_mypy exit code was only the return code of the executable and not the number or errors raised by it.

### DIFF
--- a/ament_mypy/ament_mypy/main.py
+++ b/ament_mypy/ament_mypy/main.py
@@ -104,8 +104,10 @@ def main(argv: List[str] = sys.argv[1:]) -> int:
     print('\n{} files checked'.format(len(filenames)))
     if not normal_report:
         print('No errors found')
+        exit_code = 0
     else:
         print('{} errors'.format(len(errors_parsed)))
+        exit_code = len(errors_parsed)
 
     print(normal_report)
 


ament_mypy/ament_mypy/main.py
```diff
@@ -104,8 +104,10 @@ def main(argv: List[str] = sys.argv[1:]) -> int:
     print('\n{} files checked'.format(len(filenames)))
     if not normal_report:
         print('No errors found')
+        exit_code = 0
     else:
         print('{} errors'.format(len(errors_parsed)))
+        exit_code = len(errors_parsed)
```
christophebedard, line 110: I'm not sure about this, because `errors_parsed` may not actually be errors. They could be "notes," which aren't really errors and shouldn't make the test exit with a non-zero exit code.

tomkimsour: Ah yes so my issue statement was wrong and the default behavior of mypy I think stems from this statement :

> Python will remain a dynamically typed language, and the authors have no desire to ever make type hints mandatory, even by convention

However I have a hard time seeing the purpose of mypy as an ament_package if it's not to enforce typing. As is, it is checking anything. I don't feel like it is coming with good default like other linters or formatters could have here. Would it make sense to update the default config file of mypy ?

tomkimsour:
Closing the PR , let's continue the discussion in the issue #510

PR #515, opened 2024-12-30
Improve ament_copyright performance drastically.

clalancette:
While running some repeated tests locally, I was finding that ament_copyright was taking about 3 seconds to process a small number of files (34).  After looking into it, I found that it was repeatedly calling get_licenses() and get_copyright_names().

However, we notice that get_licenses() and get_copyright_names() are pulling the known copyrights from the package itself. Thus, there is no way that these can change during the runtime of the program.

This PR changes it so that we only call these functions once in main(), and the rest of the code just uses that.  With this in place, the runtime for the 34 files I was checking went from 3 seconds down to 0.2 seconds.

### DIFF
--- a/ament_copyright/ament_copyright/parser.py
+++ b/ament_copyright/ament_copyright/parser.py
@@ -17,8 +17,6 @@ import re
 
 from ament_copyright import ALL_FILETYPES
 from ament_copyright import CONTRIBUTING_FILETYPE
-from ament_copyright import get_copyright_names
-from ament_copyright import get_licenses
 from ament_copyright import LICENSE_FILETYPE
 from ament_copyright import SOURCE_FILETYPE
 from ament_copyright import UNKNOWN_IDENTIFIER
@@ -52,14 +50,12 @@ class FileDescriptor:
         with open(self.path, 'r', encoding='utf-8') as h:
             self.content = h.read()
 
-    def parse(self):
+    def parse(self, licenses, known_copyrights):
         raise NotImplementedError()
 
-    def identify_license(self, content, license_part, licenses=None):
+    def identify_license(self, content, license_part, licenses):
         if content is None:
             return
-        if licenses is None:
-            licenses = get_licenses()
         formatted_content = remove_formatting(content)
 
         for name, license_ in licenses.items():
@@ -93,8 +89,7 @@ class SourceDescriptor(FileDescriptor):
 
         self.copyright_identifiers = []
 
-    def identify_copyright(self):
-        known_copyrights = get_copyright_names()
+    def identify_copyright(self, known_copyrights):
         for c in self.copyrights:
             found_name = c.name
             for identifier, name in known_copyrights.items():
@@ -104,7 +99,7 @@ class SourceDescriptor(FileDescriptor):
             else:
                 self.copyright_identifiers.append(UNKNOWN_IDENTIFIER)
 
-    def parse(self):
+    def parse(self, licenses, known_copyrights):
         self.read()
         if not self.content:
             return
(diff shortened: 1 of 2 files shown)


clalancette:
Pulls: https://github.com/ament/ament_lint/pull/515
Gist: https://gist.githubusercontent.com/clalancette/159262ee8c8be36bd6f46c1820a3912c/raw/bc99cff018c009c1bc8b427c8df276071086118e/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15018
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22334/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16597/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1865/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23125/)

ament_copyright/ament_copyright/parser.py
```diff
@@ -93,8 +89,7 @@ def __init__(self, path):
 
         self.copyright_identifiers = []
 
-    def identify_copyright(self):
-        known_copyrights = get_copyright_names()
+    def identify_copyright(self, known_copyrights):
```
kscottz, line 94: If I am understanding this PR correctly the crux of the change is the redundant call to `get_copyright_names`. This revision kicks that call up to the main function. 

Is that correct?

ament_copyright/ament_copyright/parser.py
```diff
@@ -104,7 +99,7 @@ def identify_copyright(self):
             else:
                 self.copyright_identifiers.append(UNKNOWN_IDENTIFIER)
 
-    def parse(self):
+    def parse(self, licenses, known_copyrights):
```
kscottz: My only concern here is that maybe, possibly, somewhere, someone is calling the `parse` function with no arguments. I doubt that is the case, but maybe we want to throw in a deprecation warning / NotImplemented error?

kscottz (approved):
I didn't see anything that jumped out at me but I am not in a position to test it the changes. From what I can gather anything this might impact is downstream, so it is worth merging and seeing if there are consequences.

ament_copyright/ament_copyright/parser.py
clalancette: That's a fair concern.  What I can do there is to make `licenses` and `known_copyright` set to `None` by default, and if they are unset, get them.  That way any "old" callers will still get the functionality (though it will be slower), and new callers will get the maximum performance.

clalancette: And now fixed in d4ed942

ament_copyright/ament_copyright/parser.py
clalancette: Yes, exactly.

clalancette:
Pulls: https://github.com/ament/ament_lint/pull/515
Gist: https://gist.githubusercontent.com/clalancette/3e9a587cad5068033925bd8d34b9490f/raw/bc99cff018c009c1bc8b427c8df276071086118e/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15021
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22337/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16600/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1868/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23128/)

clalancette:
I'm going to go ahead and merge this one with an approval, especially since it should speed up our CI times by about 5% or so.

PR #522, opened 2025-02-14
(ament_cpplint) Update cpplint version to v2.0.0

### LINKED ISSUE #500: ament_cpplint Support for C++20

EricVoll:
PR #501 already contained an update to the latest `cpplint` version at the time. It was decided to split the changes there into two PRs. This PR updates the cpplint version to `v2.0.0`, which is superseded by 5 PRs fixing various things. 

I went with v2.0.0. to have an officially released version instead of some mid-release version.

Fixes #500

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 #
 # Copyright (c) 2009 Google Inc. All rights reserved.
 #
@@ -28,8 +28,6 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# https://github.com/cpplint/cpplint/blob/6b1d29874dc5d7c3c9201b70e760b3eb9468a60d/cpplint.py
-
 """Does google-lint on c++ files.
 
 The goal of this script is to identify places in the code that *may*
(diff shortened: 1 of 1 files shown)


christophebedard (changes requested):
Thanks for the PR. Our policy is to use the version available from system packages on Ubuntu. This way the version is consistent across all target platforms. Ubuntu 24.04 Noble (the Ubuntu version targeted for Rolling and Jazzy) currently provides version 1.6.1, so we could bump to that: https://packages.ubuntu.com/noble/cpplint.

(The current version is 1.5.5 because that was the version provided by Ubuntu 22.04 Jammy, and it just hasn't been bumped yet: https://packages.ubuntu.com/jammy/cpplint)

Also, please sign off your commit, see: https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#change-control-process.

christophebedard:
> which is superseded by 5 PRs fixing various things.

I wonder if we should re-apply the changes that we made on top: https://github.com/ament/ament_lint/commits/185613c26a3f65cc802dc5bc05d42fd36381040f/ament_cpplint/ament_cpplint/cpplint.py

christophebedard:
> Ubuntu 24.04 Noble (the Ubuntu version targeted for Rolling and Jazzy) currently provides version 1.6.1, so we could bump to that: https://packages.ubuntu.com/noble/cpplint.

#562 does this, and there's been no response on this PR for some time, so I'll close this.

Issue #529, opened 2025-03-19
Improve Repository Documentation


Nils-ChristianIseke:
I wanted to contribute to the repository but found it challenging due to the lack of a CONTRIBUTING.md guide and a detailed README. Adding clear contribution guidelines and a more comprehensive README would make the process more accessible for new contributors and clarify the repository's purpose and scope.

Additionally, I noticed the absence of a CI pipeline for linting checks. Is there a specific reason for not supporting this? Or am i missing something?

I’d be happy to help with these improvements!

christophebedard:
Great to hear that you want to contribute! You can find general contribution guidelines here: https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing.html. If you have any specific questions, feel free to ask.

We should probably add _something_ in the README, though. It could start with an overview of the repo and its packages.

> Additionally, I noticed the absence of a CI pipeline for linting checks. Is there a specific reason for not supporting this? Or am i missing something?

We use Jenkins CI on ci.ros2.org. Maintainers can trigger jobs for PRs.

Nils-ChristianIseke:
After reading the documentation and some discussions on ros-discourse, i think there are two main reasons to use ament_lint:
1. Colcon support for the tools
2. Enforcing default tooling configurations across ROS2 Packages.

However, for many of the ament_lint tools you can use your own configurations, this clashes with the 2nd reason. Wouldn't it be better if the ament_tools are completely unconfigurable regarding the linting rules?

christophebedard:
For the "core" ROS 2 packages, we indeed want to have the same configuration, e.g., https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html.

However, ROS is a federated community with a ton of packages. It makes sense for authors of those packages (which may or may not get released or even ever be run on any other computer) to use these linting tools and to let them configure them however they like.

Nils-ChristianIseke:
Thank you. And sry for asking this noob questions :).
Just to clarify: **This isn’t meant to be criticism—it’s a request to help  me (and hopefully others) to align with the project’s goals.**
Do you think it is reasonable to add a README containing a little bit of information?
I created a PR (#533) which is far from perfect, feedback is very much appreciated.

christophebedard:
[...]

PR #533, opened 2025-03-21
DRAFT README.md

Nils-ChristianIseke:
Regarding #529.
This is just a first draft for a README (Far from beeing perfect).
I would appreciate some feedback regarding the points below. 

This PR proposes an initial README to make the repository more self-contained. While minimal, it aims to:

- Clarify ament_lint’s purpose for new contributors/users.
- Explain key benefits /shortcoming over manual linting setups.
- Provide quick-start examples for common workflows (e.g. by refering the tools docs)

**I still have the following open questions which might be answered by the readme**:

- Why Not Just Share Tool Configs?
If the goal is to standardize linting, why bundle configurations with CMake integration instead of distributing standalone configs (e.g., .clang-format, flake8.ini) that developers could adopt via other tools like pre-commit?

- Abstraction Tradeoffs
While CMake integration benefits some workflows, it adds a layer that complicates debugging or customization for others. For example, developers using pre-commit might see ament_lint as more of a burdon.

- Could ament_lint focus on integration while decoupling from configuration?

- ROS Core Team’s Reasoning, e.g:
Ament_lint is the officially supported ROS 2 solution for linting because it ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,14 @@
-# Ament Lint
-
-ament_lint integrates linting into ROS 2’s build system, automating code quality checks based on [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+ament_lint integrates linting tools into ROS 2’s build system, automating code quality checks based on the [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+It bundles common linting tools with ROS 2-specific wrappers to ensure consistent functionality across packages.
 
 ## Key Benefits
+- Seamless Build Integration – No manual scripting for tool setup or result parsing.
+- Preconfigured Standards – Enforces ROS 2 guidelines for C++, Python, XML, etc.
 
-- Build Integration
-- Linting runs during colcon build
-- No manual scripting for tool setup or result parsing.
+## Contributing
 
-## ROS 2 Standards
-
-Preconfigured rules to match [ros2 coding guidelines](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html) for C++, Python, XML, and CMake.
-
-## Flexibility
-
-If you don't want to stick to the ros2 coding guidelines use your own configurations (e.g., ament_cppcheck(ARGS "...")) while retaining CMake integration.
-
-## Contribution
-
-In particular, the configuration of the tools is not yet perfect and not everything you are used to from the basic tools works. Feel free to contribute to this project and improve the configuration options
-Feel free two contribute to this repo, please follow the [Contrubution Guidelines](CONTRIBUTING.md).
+The tool argument are still being refined — some features from standalone linters may not yet work perfectly. We welcome contributions to improve:
+- Linter configurations
+- Additional tool support
+- ...
+Please follow the [Contrubution Guidelines](CONTRIBUTING.md).
--- a/README.md
+++ b/README.md
@@ -1,0 +1,22 @@
+# Ament Lint
+
+ament_lint integrates linting into ROS 2’s build system, automating code quality checks based on [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+
+## Key Benefits
+
+- Build Integration
+- Linting runs during colcon build
+- No manual scripting for tool setup or result parsing.
+
+## ROS 2 Standards
+
+Preconfigured rules to match [ros2 coding guidelines](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html) for C++, Python, XML, and CMake.
+
+## Flexibility
+
+If you don't want to stick to the ros2 coding guidelines use your own configurations (e.g., ament_cppcheck(ARGS "...")) while retaining CMake integration.
+
+## Contribution
+
+In particular, the configuration of the tools is not yet perfect and not everything you are used to from the basic tools works. Feel free to contribute to this project and improve the configuration options
+Feel free two contribute to this repo, please follow the [Contrubution Guidelines](CONTRIBUTING.md).
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-ament_lint integrates linting tools into ROS 2’s build system, automating code quality checks based on the [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
+The tools are configured to meet [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
 It bundles common linting tools with ROS 2-specific wrappers to ensure consistent functionality across packages.
 
 ## Key Benefits
--- a/README.md
+++ b/README.md
@@ -1,11 +1,17 @@
-ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
+## ament_lint
+
+ament_lint contains a number of packages, each package wraps a linter into an ament_$linter CLI tool, which is then wrapped in CMake (ament_cmake_$linter)
+Those linters can be used both from the command line and from CMake.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,11 @@
+ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
+The tools are configured to meet [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+It bundles common linting tools with ROS 2-specific wrappers to ensure consistent functionality across packages.
```
christophebedard: You could link to this very relevant tutorial: https://docs.ros.org/en/rolling/Tutorials/Advanced/Ament-Lint-For-Clean-Code.html

Also, I think all `ament_*` tools (and their `ament_cmake_*` wrapper) have a short documentation file, e.g., https://github.com/ament/ament_lint/blob/rolling/ament_mypy/doc/index.rst. You could refer to it, for example:

> For information on each tool, refer to the documentation in their package's `doc/` directory.

README.md
```diff
@@ -0,0 +1,11 @@
+ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
```
christophebedard: FYI we usually try to only have one sentence per line in documentation-related documents. This minimizes the git diff when we make changes, which makes maintenance easier.

So for this line (and some of the lines below):

```suggestion
ament_lint contains a number of packages that perform linting tasks.
Those linters can be used both from the command line and from CMake.
```

README.md
```diff
@@ -0,0 +1,11 @@
...
+The tools are configured to meet [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+It bundles common linting tools with ROS 2-specific wrappers to ensure consistent functionality across packages.
+
+## Contributing
+
+The tool argument are still being refined — some features from standalone linters may not yet work perfectly. We welcome contributions to improve:
+- Linter arguments
+- Additional tool support
+- ...
+Please follow the [Contrubution Guidelines](CONTRIBUTING.md).
```
christophebedard:
```suggestion
Please follow the [Contribution Guidelines](CONTRIBUTING.md).
```

README.md
```diff
@@ -0,0 +1,11 @@
+ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
+The tools are configured to meet [ROS 2 coding standards](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html).
+It bundles common linting tools with ROS 2-specific wrappers to ensure consistent functionality across packages.
+
+## Contributing
+
+The tool argument are still being refined — some features from standalone linters may not yet work perfectly. We welcome contributions to improve:
+- Linter arguments
+- Additional tool support
+- ...
```
christophebedard: I wouldn't include the first sentence, since that's pretty much a given :grin:

```suggestion
We welcome contributions to improve:
- Linter arguments
- Additional tool support
- ...

```

christophebedard:
I don't want to pretend that I know all the answers, but I can try to share what I know.

> Provide quick-start examples for common workflows (e.g. by refering the tools docs)

You could just point to each tool's documentation, see my other comment.

> * Why Not Just Share Tool Configs?
>   If the goal is to standardize linting, why bundle configurations with CMake integration instead of distributing standalone configs (e.g., .clang-format, flake8.ini) that developers could adopt via other tools like pre-commit?

I think it's just easier, especially for the ROS 2 core. And, for other packages, I think the tools matter more than their default configuration file, since that often depends on preferences.

Also, you can run these tools yourself through the CLI, so they _can_ be integrated into pre-commit hooks.

> * Abstraction Tradeoffs
>   While CMake integration benefits some workflows, it adds a layer that complicates debugging or customization for others. For example, developers using pre-commit might see ament_lint as more of a burden.

With `ament_lint` tools, the linter is wrapped into an `ament_$linter` CLI tool, which is then wrapped in CMake (`ament_cmake_$linter`) or called in a minimal Python file that then becomes a "normal" test, for example: https://github.com/ros2/ros2_tracing/blob/rolling/tracetools_trace/test/test_mypy.py. It is definitely just _one_ approach, and other people have also pointed this out before: linters are usually separate from tests. And using standalone linters often makes setting up/maintaining pre-commit hooks easier.

> * Could ament_lint focus on integration while decoupling from configuration?

Possibly, but there hasn't been a good reason to do so yet, or at least a reason that would make the additional complexity/maintenance worth it.

> * ROS Core Team’s Reasoning, e.g:
>   Ament_lint is the officially supported ROS 2 solution for linting because it ...

The answer to this is probably partly historical. (As an aside, you might wonder: "why is it called "ament?" It's called ament because the ROS 1 equivalent(ish) of this toolset/buildsystem layer was called catkin, another name for the same [flower](https://en.wikipedia.org/wiki/Catkin). But why was _that_ called "catkin?" It goes on and on :smile:)

I wouldn't necessarily say it's the "officially supported ROS 2 solution for lining;" it's just the most common/most well-integrated linting solution we have. As for why that is, I'd say ROS 2 converged to this solution because of what I mentioned above. Overall, given that ROS 2 is made up of a bunch of repos, it's easier to set up linters as normal tests instead of commit hooks.

README.md
```diff
@@ -1,11 +1,17 @@
-ament_lint contains a number of packages that perform linting tasks. Those linters can be used both from the command line and from CMake.
+## ament_lint
+
+ament_lint contains a number of packages, each package wraps a linter into an ament_$linter CLI tool, which is then wrapped in CMake (ament_cmake_$linter)
+Those linters can be used both from the command line and from CMake.
```
christophebedard, line 4: They also have a Python API, e.g., here: https://github.com/ament/ament_lint/blob/2bee9151aba25b94020241c00f468f12944ab265/ament_mypy/ament_mypy/main.py#L30. This can be called in a test, e.g., here: https://github.com/ros2/launch/blob/f8ef67a0f8cface863e1cd168f0b6bf2976be60b/launch/test/test_mypy.py#L22.

Nils-ChristianIseke:
Thanks for this detailed reply :). 

I'm struggling a bit to write a good summary for these thoughts.
  E.g. something like: Ament_lint is for you if you want xyz, you better not use ament_lint if you want xyz. Do you have a suggestion?

I think it's mainly because I don't understand the point of ament_lint. I understand that it's convenient for some developers to have the cmake integration, but I think many developers would also like the ability to apply the Ros core coding styles via linters without being forced to maintain their own configuration and/or use ament_lint. Personally, for example, I would just use pre-commit and run the pure linters. Maybe we can think about providing a pre-commit configruation that ros2 users can use as alternatives to the wrapped linters?. Am I missing something with this approach?

Nils-ChristianIseke:
@christophebedard friendly ping :).

christophebedard:
I don't think there's any _deep_ reason for the way it is. It's just one solution, and it's the one we have. It's used by the ROS 2 core, and it can be used by other users if they want to use it.

In some cases, the `ament_[cmake_]*` wrapper also makes sure to create a standard xUnit result file:

* https://github.com/ament/ament_lint/blob/1e737ea8d94a5207eaaecdfe640c4805750a350a/ament_cmake_uncrustify/cmake/ament_uncrustify.cmake#L54-L55
* https://github.com/ament/ament_lint/blob/1e737ea8d94a5207eaaecdfe640c4805750a350a/ament_uncrustify/ament_uncrustify/main.py#L216-L233

but, again, this is because these tools are primarily used through _tests_ (see `test_mypy.py` in my comment above), which a lot of people don't like, even if you can run `ament_mypy` directly in a terminal ([or in CI](https://github.com/ros-tooling/action-ros-lint)). You could mention that.

> Maybe we can think about providing a pre-commit configruation that ros2 users can use as alternatives to the wrapped linters?. Am I missing something with this approach?

I think that's probably outside of the scope of this repo, but people can do it if they want (and I know some have done it). One reason it's not done for the ROS 2 core itself is that there are _so many_ repos and it would be annoying to set up hooks and maintain them.

> but I think many developers would also like the ability to apply the Ros core coding styles via linters without being forced to maintain their own configuration and/or use ament_lint.

I mean, you could just use the configuration file used by the corresponding `ament_*` tool, e.g.:

* `ament_flake8`: https://github.com/ament/ament_lint/blob/rolling/ament_flake8/ament_flake8/configuration/ament_flake8.ini
* `ament_mypy`: https://github.com/ament/ament_lint/blob/rolling/ament_mypy/ament_mypy/configuration/ament_mypy.ini
* `ament_uncrustify`: https://github.com/ament/ament_lint/blob/rolling/ament_uncrustify/ament_uncrustify/configuration/ament_code_style_0_78.cfg

You could fetch these via `curl` or whatever.

PR #548, opened 2025-07-18
Update cpplint

mosfet80:
Update cpplint from version 1.5.5. to 2.0.2
changelog: https://github.com/cpplint/cpplint/releases


christophebedard:
We use the version of cpplint available from the Ubuntu system packages. For Ubuntu 22.04, that is version 1.5.5: https://packages.ubuntu.com/jammy/cpplint. For Ubuntu 24.04, that is version 1.6.1: https://packages.ubuntu.com/noble/cpplint. A bump to 1.6.1 is proposed in #562, so I'll close this PR.

PR #32, opened 2015-10-15
allow C-style casts in c code

dirk-thomas:

PR #562, opened 2025-10-18
Bumps cpplint to 1.6.1

InvincibleRMC:
Bumps cpplint to the current version on Ubuntu 24.04.

Current version info found [here](https://pkgs.org/search/?q=cpplint).

If `rhel` gets cpplint could maybe stop vendoring `cpplint.py` in the future.

Relate issues:
#548
#522
#501
#500

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -5877,6 +5877,10 @@ def CheckCStyleCast(filename, clean_lines, linenum, cast_type, pattern, error):
   if Match(r'^\s*(?:;|const\b|throw\b|final\b|override\b|[=>{),]|->)',
            remainder):
     return False
+  
+  # Don't warn in C files about C-style casts
+  if os.path.splitext(filename)[1] in ['.c', '.h']:
+    return False
 
   # At this point, all that should be left is actual casts.
   error(filename, linenum, 'readability/casting', 4,
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -28,7 +28,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# https://github.com/cpplint/cpplint/blob/6b1d29874dc5d7c3c9201b70e760b3eb9468a60d/cpplint.py
+# https://github.com/cpplint/cpplint/blob/1.6.1/cpplint.py
 
 """Does google-lint on c++ files.
 
@@ -42,6 +42,11 @@ In particular, we can get very confused by /* and // inside strings!
 We do a small hack, which is to ignore //'s with "'s after them on the
 same line, but it is far from perfect (in either direction).
 """
+
+# cpplint predates fstrings
+# pylint: disable=consider-using-f-string
+
+# pylint: disable=invalid-name
 
 import codecs
 import copy
@@ -61,13 +66,15 @@ import xml.etree.ElementTree
 # if empty, use defaults
 _valid_extensions = set([])
 
-__VERSION__ = '1.5.5'
+__VERSION__ = '1.6.1'
 
 try:
+  #  -- pylint: disable=used-before-assignment
   xrange          # Python 2
 except NameError:
   #  -- pylint: disable=redefined-builtin
   xrange = range  # Python 3
+
 
 _USAGE = """
 Syntax: cpplint.py [--verbose=#] [--output=emacs|eclipse|vs7|junit|sed|gsed]
@@ -879,12 +886,14 @@ _line_length = 80
 _include_order = "default"
 
 try:
+  #  -- pylint: disable=used-before-assignment
   unicode
 except NameError:
   #  -- pylint: disable=redefined-builtin
   basestring = unicode = str
 
 try:
+  #  -- pylint: disable=used-before-assignment
   long
 except NameError:
   #  -- pylint: disable=redefined-builtin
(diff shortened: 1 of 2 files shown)


christophebedard:
Pulls: ament/ament_lint#562
Gist: https://gist.githubusercontent.com/christophebedard/c2878c607000610bb34990690e9fa862/raw/905fb39315abed5400bd906b1397e52746aec6f0/ros2.repos
BUILD args:  --packages-above-and-dependencies ament_cpplint
TEST args:  --packages-above ament_cpplint
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17314
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25453/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19538/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4988/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25752/)

christophebedard:
Let's see if this breaks any tests downstream. If it does, we'll have to fix them at the same time.

InvincibleRMC:
There seems to be an oversight in `cpplint 1.6.1` that it is running `C++` lints on `C` files. To disable this you add `// LINT_C_FILE` to every `.c` file. This has already been adjusted upstream to run checks base on the `.c` file extensions as well.

https://github.com/cpplint/cpplint/blob/b728b3f2b013b8989bd626361ab21ae85f222830/cpplint.py#L1207-L1211

I imagine taking the patch would be less disruptive than adding `LINT_C_FILE` to every c file in ROS.

ament_cpplint/ament_cpplint/cpplint.py
```diff
@@ -5882,10 +5878,6 @@ def CheckCStyleCast(filename, clean_lines, linenum, cast_type, pattern, error):
            remainder):
     return False
 
-  # Don't warn in C files about C-style casts
```
InvincibleRMC, line 5885: Oh I see this `patch`  already existed but was not documented.

InvincibleRMC: For reference added 9 years ago. :smiley: https://github.com/ament/ament_lint/pull/32

christophebedard:
Pulls: ament/ament_lint#562
Gist: https://gist.githubusercontent.com/christophebedard/181b266114b97d4e6455eac316112568/raw/905fb39315abed5400bd906b1397e52746aec6f0/ros2.repos
BUILD args:  --packages-above-and-dependencies ament_cpplint
TEST args:  --packages-above ament_cpplint
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17326
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25466/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19552/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5001/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25766/)

ament_cpplint/ament_cpplint/cpplint.py
```diff
@@ -5332,28 +5343,12 @@ def CheckLanguage(filename, clean_lines, linenum, file_extension,
           'Did you mean "memset(%s, 0, %s)"?'
           % (match.group(1), match.group(2)))
 
-  # Check for 'using namespace' which pollutes namespaces.
```
InvincibleRMC, line 5335: Hmmm. It seems this was another un-documented patch see https://github.com/ament/ament_lint/pull/67. One the one hand it has rejected by the upstream google maintainers however, we could maybe try again with the cpplint fork. On the other hand it seemed to be more of work around of [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65923) gcc bug that has been fixed for some time now.

christophebedard: I would re-apply the patch just to avoid all those linter errors. We can decide whether to drop it (and fix downstream packages) separately.

maspe36:
Any movement on this? I enjoy using `ament_cpplint()` to run cpplint w/ ctest, but I need to `NOLINT` a few C++20 features, notably concepts. 

Also, with [gcc14 being shipped in 24.04](https://packages.ubuntu.com/noble/gcc-14) (and [gcc15+ in 26.04](https://code.launchpad.net/ubuntu/resolute/+package/gcc-15)) I will start using C++23 which may exacerbate things.

InvincibleRMC:
This PR stalled out since I got distracted working on other projects. I might have some time in a month or 2 to revisit. However the problems around concepts would need to wait for Ubuntu 26 and cpplint 2.0+. Which will likely be an even larger effort than just this minor bump.

InvincibleRMC:
This PR stalled out since I got distracted working on other projects. I might have some time in a month or 2 to revisit. However the problems around concepts would need to wait for Ubuntu 26 and cpplint 2.0+. Which will likely be an even larger effort than just this minor bump.

christophebedard:
> However the problems around concepts would need to wait for Ubuntu 26 and cpplint 2.0+. Which will likely be an even larger effort than just this minor bump.

We just talked about this during the triage meeting :waffle:. We should indeed work on moving to version 2.0. Ubuntu 26.04 Resolute has 2.0.2: https://packages.ubuntu.com/resolute/cpplint. We should start by bumping to that version to get an idea of the impact.

InvincibleRMC:
>
> We just talked about this during the triage meeting :waffle:. We should indeed work on moving to version 2.0. Ubuntu 26.04 Resolute has 2.0.2: https://packages.ubuntu.com/resolute/cpplint. We should start by bumping to that version to get an idea of the impact.

Are you saying it would be fine to jump straight to 2.0.2?

christophebedard:
For Rolling, yes. Ideally, we would bump + fix everything at the same time. However, if the version bump breaks all packages (i.e., makes all their cpplint test fail and creates a ton of work), then we can come up with a different plan, e.g., bump and then do more of a gradual upgrade separately (?).

PR #569, opened 2026-01-25
Add `ament_lint_auto_py` and `ament_lint_common_py`

InvincibleRMC:
## Description

Adding `ament_lint_auto_py` and `ament_lint_common_py` to allow auto linting for `ament_python` packages. It should have 1 to 1 parity with `ament_lint_auto` features. Including support for `ament_lint_auto_exclude` and `ament_lint_auto_file_exclude`.

### Is this user-facing behavior change?

Yes it would simpler running linters in `ament_python` build packages. Throughout the core sometimes linters get missed. And if we ever wanted to add more in the future [like `ament_mypy` for typing ;)] being able to use similar logic to `ament_lint_auto` in `CMakeLists.txt` would be very useful.


peci1:
I very much like this initiative.

Does it mean that all the copy-pasted lint tests in downstream repos would be replaced just by something like test_ament_lint_auto.py from the readme? Is there no way to do it even without this one (e.g. generate it at build time)?

It's not clear to me why does ament_lint_common_py contain also the C++-only linters like cppcheck. Does it make sense in some scenarios? Or is it just to have 1:1 mapping between cpp and py linters? Regardless of the answer, I think this should be documented in the readme.

Also, setting excludes only via environment variables is insufficient. That's good for local development, but unusable on the buildfarm and in other automated build environments. Or am I missing something?

If specifying the excludes in setup.py is not possible, what about the export section in package.xml? But I don't like this solution either because it would leak implementation details like file names leak into the overall metadata provided by package.xml.

InvincibleRMC:
Do to your questions/feedback I have done some updates.

1. I was able to remove needing the `def test_ament_lint_auto()` function from being called downstream now just having `ament_lint_auto_py` in the `package.xml` will cause all linters listed in the `package.xml` to run.
2. Added a note about keeping parity between `ament_lint_common` and `ament_lint_common_py`. I could see a world we remove the C++ lint dependencies but, the initial implementation goal was parity.
3. For settings I moved from environment variables to pytest variables to be configured in a `pytest.ini` /`pyproject.toml` per package.

InvincibleRMC:
@fujitatomoya would you mind reviewing this?

ament_lint_auto_py/ament_lint_auto_py/pytest_plugin.py
```diff
@@ -0,0 +1,120 @@
...
+        return  # Package does not opt-in
+
+    effective_depends = get_depends_recursive(pkg_xml)
+    linters = entry_points(group='ament_lint')
+
+    for ep in linters:
+        runner = ep.load()
+
+        if runner.NAME not in effective_depends:
+            continue  # skipping linter if not declared in depends of a package.xmk
```
fujitatomoya: typo

```suggestion
            continue  # skipping linter if not declared in depends of a package.xml
```

ament_lint_auto_py/doc/index.rst
```diff
@@ -0,0 +1,107 @@
...
+---------------------------------------------------------
+
+To register a third party linter implement class like the following.
+
+.. code:: python
+
+  class CustomRunner:
+
+      NAME = 'ament_custom'
+      FILE_TYPES = ('*.py')
```
fujitatomoya: this should be tuple, not string, right?

```suggestion
      FILE_TYPES = ('*.py',)
```

ament_pep257/ament_pep257/main.py
```diff
@@ -55,6 +56,18 @@
 ]
 
 
+class Pep257Runner:
+
+    NAME = 'ament_pep257'
+    FILE_TYPES = ('*.py')
```
fujitatomoya:
```suggestion
    FILE_TYPES = ('*.py',)
```

ament_pycodestyle/ament_pycodestyle/main.py
```diff
@@ -17,12 +17,26 @@
...
 from xml.sax.saxutils import escape
 from xml.sax.saxutils import quoteattr
 
 import pycodestyle
 
 
+class PycodestyleRunner:
+
+    NAME = 'ament_pycodestyle'
+    FILE_TYPES = ('*.py')
```
fujitatomoya:
```suggestion
    FILE_TYPES = ('*.py',)
```

ament_pyflakes/ament_pyflakes/main.py
```diff
@@ -26,6 +27,19 @@
 from pyflakes.reporter import Reporter
 
 
+class PyflakesRunner:
+
+    NAME = 'ament_pyflakes'
+    FILE_TYPES = ('*.py')
```
fujitatomoya:
```suggestion
    FILE_TYPES = ('*.py',)
```

ament_xmllint/ament_xmllint/main.py
```diff
@@ -28,7 +29,20 @@
 from xml.sax.saxutils import quoteattr
 
 
-def main(argv=sys.argv[1:]):
+class XmlLintRunner:
+
+    NAME = 'ament_xmllint'
+    FILE_TYPES = ('*.xml')
```
fujitatomoya:
```suggestion
    FILE_TYPES = ('*.xml',)
```

ament_lint_auto_py/ament_lint_auto_py/pytest_plugin.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if runner.NAME in excluded:
+            continue  # skipping linter if declared in ament_lint_auto_exclude
+
+        # skip linters if no matching files exist
+        found_file = False
+        for pattern in runner.FILE_TYPES:
+            if any(Path('.').rglob(pattern)):
```
fujitatomoya: probably using `config.rootpath`, instead of CWD?

```suggestion
            if any(config.rootpath.rglob(pattern)):
```

ament_lint_auto_py/doc/index.rst
```diff
@@ -0,0 +1,107 @@
+ament_lint_auto_py
+===============
```
fujitatomoya:
```suggestion
ament_lint_auto_py
==================
```

ament_lint_auto_py/doc/index.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code:: xml
+
+    <test_depend>ament_lint_auto_py</test_depend>
+
+    <!-- this recursively depends on a set of common linters -->
+    <test_depend>ament_lint_common_py</test_depend>
+
+
+How to exclude linter modules with ament_lint_auto_py?
+---------------------------------------------------
```
fujitatomoya:
```suggestion
How to exclude linter modules with ament_lint_auto_py?
------------------------------------------------------
```

ament_lint_auto_py/doc/index.rst
```diff
@@ -0,0 +1,107 @@
...
+As an example to exclude the `copyright` linter:
+
+.. code::
+    [pytest]
+
+    ament_lint_auto_exclude = ament_copyright
+
+
+How to exclude files with ament_lint_auto_py?
+------------------------------------------
```
fujitatomoya:
```suggestion
How to exclude files with ament_lint_auto_py?
---------------------------------------------
```

ament_lint_common_py/doc/index.rst
```diff
@@ -0,0 +1,24 @@
+ament_lint_common_py
+=================
```
fujitatomoya:
```suggestion
ament_lint_common_py
====================
```

fujitatomoya (changes requested):
i got several minor comments only, i would like someone else who is familiar with ament to review this before merge.

PR #572, opened 2026-01-31
Enable quiet mode for cpplint (backport Jazzy)

mini-1235:
https://github.com/ament/ament_lint/pull/571 

cc @christophebedard

### DIFF
--- a/ament_cpplint/ament_cpplint/main.py
+++ b/ament_cpplint/ament_cpplint/main.py
@@ -158,6 +158,8 @@ def main(argv=sys.argv[1:]):
         else:
             if not args.quiet:
                 print("Not using '--root'")
+        if not args.quiet:
+            print('')
 
         arguments += files
         filenames = ParseArguments(arguments)
--- a/ament_cpplint/ament_cpplint/main.py
+++ b/ament_cpplint/ament_cpplint/main.py
@@ -181,6 +181,8 @@ def main(argv=sys.argv[1:]):
 
             ProcessFile(filename, _cpplint_state.verbose_level)
             report.append((filename, errors))
+            if errors or not args.quiet:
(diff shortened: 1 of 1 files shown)


ament_cpplint/ament_cpplint/main.py
```diff
@@ -163,9 +153,11 @@ def main(argv=sys.argv[1:]):
...
             arguments.append(root_arg)
-            LOGGER.debug("Using '%s' argument", root_arg)
+            if not args.quiet:
+                print("Using '%s' argument" % root_arg)
         else:
-            LOGGER.debug("Not using '--root'")
+            if not args.quiet:
+                print("Not using '--root'")
 
         arguments += files
```
christophebedard, line 164: There was a `print('')` (although `print()` would be equivalent) call here originally. I'd prefer to keep it in older distros to keep the exact same behaviour, although we should add an `if not args.quiet:` around it to match the prints above:

```suggestion
                print("Not using '--root'")
        if not args.quiet:
            print('')

        arguments += files
```

ament_cpplint/ament_cpplint/main.py
christophebedard, line 183: I would keep the print right after this line too. If we're in quiet mode, I assume the above code will only print something if there are errors, so we'd need something like this:

```py
            if errors or not args.quiet:
                print('')
```

(for some reason, creating an actual suggestion here gets all messed up unfortunately)

christophebedard (commented):
Thanks, I have some minor suggestions

mini-1235:
Thanks for the review @christophebedard! I have updated the code accordingly

christophebedard (approved):
Thanks! I'll merge this into the backport PR and run CI there.

Issue #497, opened 2024-07-25
ament_pep257 uses deprecated pydocstyle


InvincibleRMC:
It currently still works but has been [deprecated](https://github.com/PyCQA/pydocstyle/pull/658) in favor of [ruff](https://github.com/astral-sh/ruff). However as of https://github.com/astral-sh/ruff/issues/659 there is not currently a Python api that could be used to replace it. And even if it did work ruff is not available via package managers yet.

Some possible solutions are the following: 
1. Add warning into ament_pep257 that it might break in the future
2.  Hold out for a ruff api replacement that could be dropped in pydocstyles' place in the future
3.  Add a deprecation warning then in some future version deprecate in favor of some future ament_ruff

mjcarroll:
While `ruff` is interesting, as you mentioned there isn't a Python API that makes it readily available to be used.

When discussing in the weekly maintenance triage, it was also pointed out that `flake8` also likely covers most of what we need from pep257.

In the short term, a pull request adding a deprecation warning for `ament_pep257` (option 3) would be acceptable.  Thanks for bringing this up.

Nils-ChristianIseke:
@mjcarroll do we really need the python api to make this work? Calling ruff via a subprocess is possible as well (see #531).

mjcarroll:
> [@mjcarroll[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/mjcarroll@google.com)](https://github.com/mjcarroll) do we really need the python api to make this work? Calling ruff via a subprocess is possible as well (see [#531](https://github.com/ament/ament_lint/pull/531)).

Python API isn't strictly necessary in my mind.  If someone was to develop a parallel linter integration (`ament_ruff`) that could be a drop-in replacement.

InvincibleRMC:
If we wanted to add `ament_ruff` I know there is a custom ament_ruff out there. We could go the  subprocess route. And now that we have cargo we could in theory use the rust api and make our bindings just for ament.

InvincibleRMC:
@mjcarroll thoughts on stuff getting more rusty? If sounds ok I might spend this week or 2 creating a prototype.

InvincibleRMC:
[...]

PR #578, opened 2026-03-26
[ament_pep257] Add ruff implementation 

InvincibleRMC:
Could be used on `rhel 10` where `pydocstyle` doesn't exist. Could maybe speed up windows CI a bit.

Would help with #497 at least on `rhel` and windows. But still a potential problem in the future with Ubuntu if `pydocstyle` is removed before `ruff` or some alternative is added.


InvincibleRMC:
@mjcarroll thoughts on using something like this to solve no `pydocstyle` on `rhel 10`?

PR #579, opened 2026-04-22
Skip pydocstyle tests if it can't be imported

cottsay:
This change is modeled after ament_cppcheck, which is also skipped on some platforms. Unlike that change, however, we need to handle invocation of ament_pep257 from downstream pytests, so I implemented that using a SkipTest exception and an additional wrapper around `main()` so that direct ament_pep257 command line invocations still look okay.

The changes to `get_xunit_output` and addition of `write_xunit_output` were taken from ament_cppcheck.

This is intended to be a sort of stopgap to get us through the Lyrical release until something more robust like #578 can safely land.

### DIFF
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -39,12 +39,12 @@ _ament_ignore = [
 ]
 
 
-def main_with_catch(*args, **kwargs) -> Literal[0, 1, 188]:
+def main_with_catch(*args, **kwargs) -> Literal[0, 1]:
     try:
         return main(*args, **kwargs)
     except SkipTest as e:
         print(str(e), file=sys.stderr)
-        return 188
+        return 1
 
 
 def main(argv: list[str] = sys.argv[1:]) -> Literal[0, 1]:
--- a/ament_pep257/ament_pep257/main.py
+++ b/ament_pep257/ament_pep257/main.py
@@ -20,26 +20,9 @@ import os
 import sys
 import time
 from typing import Literal
+from unittest import SkipTest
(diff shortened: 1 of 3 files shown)


mjcarroll (approved):

ament_pep257/ament_pep257/main.py
```diff
@@ -56,7 +39,24 @@
 ]
 
 
+def main_with_catch(*args, **kwargs) -> Literal[0, 1, 188]:
```
InvincibleRMC: Will this not require all downstream users that call main directly have to update to this that want to support `rhel`? I know this will work for `ament_cmake_pep257`.

InvincibleRMC (approved):

ament_pep257/ament_pep257/main.py
cottsay: I dropped the explicit return code in 514734b. It's kinda misleading - it isn't actually needed to make any of our existing mechanisms work.

The downstream pytest tests for `ament_pep257` will get skipped due to the uncaught `SkipTest` exception coming out of `main`, and the `ament_cmake_pep257` will see the `ctest` as successful but the xunit output will show that everything was skipped, which is the same behavior as `ament_cppcheck`.

cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28834/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21645/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8689/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27706/)

PR #581, opened 2026-05-06
Update uncrustify version selection to support >= 0.78.1

mjcarroll:
This PR updates the Uncrustify version selection logic in `ament_uncrustify` to support versions `>= 0.78.1`.

### The Problem

Currently, `ament_uncrustify` hardcodes a check for **exactly** `0.78.1`. If the version differs (even if it is newer), the script falls back to the obsolete `0.72` configuration (designed for RHEL 8 / Ubuntu 20.04).

*   **RHEL 10** (via EPEL) provides Uncrustify **0.80.1**. 
*   Because `0.80.1 != 0.78.1`, RHEL 10 currently falls back to the `0.72` configuration, resulting in thousands of formatting violations in CI.
*   **Ubuntu 26.04** (Resolute) currently provides **0.78.1**, so it passes

### The Fix

Updated the logic to use the `0.78` configuration for any version `>= 0.78.1`. This ensures that the modern style definition is used for the 0.8x family of engines available on RHEL 10 and newer Ubuntu releases.

Assisted-by: Gemini CLI:2.0-Flash


mjcarroll:
Test run against RHEL 10 [Build Status](https://ci.ros2.org/job/ci_linux-rhel/8834/)

mjcarroll:
Keeping this as a draft, I'm kind of just hoping that newer config files will be fine then we can figure out what we actually want to do here.

mjcarroll:
Testing with this and the launch fix from here: https://github.com/ros2/launch/pull/974

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21771/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27850/)

clalancette (approved):

clalancette:
Successful CI for this is in https://github.com/ros2/launch/pull/974#issuecomment-4397708364

clalancette:
@Mergifyio backport lyrical

mjcarroll:
Ah, @cottsay recommended 

```
version = b'0.78.1'
version_tuple = tuple(int(x) for x in itertools.takewhile(bytes.isdigit, version.split(b'.')))
```

I will do that in a follow-up

PR #587, opened 2026-08-20
Update cpplint 2.0.2

mini-1235:
Hi! I have been following the tooling used in the ROS 2 core repositories in my own project, including `cpplint`. Recently, I noticed that some C++20 headers are misclassified by the version we currently use. Since ROS 2 switched to C++20 starting from Lyrical, I think it would be important for `cpplint` to support these headers properly as well.

I also noticed that some core repositories currently add `NOLINT` when including C++20 headers, for example:

https://github.com/ros/resource_retriever/blob/1137e21d1820f6e176e140eb7647a9c05626e2b2/resource_retriever/src/plugins/curl_retriever.cpp#L35-L37

IMO, though, this is not a good long-term solution.

So I asked my agent to help bump the bundled `cpplint` version to 2.0.2, which is also the version currently packaged in Ubuntu Resolute:

https://packages.ubuntu.com/search?keywords=cpplint&searchon=names&suite=resolute&section=all

## Summary

- update the vendored cpplint implementation from 1.5.5 to upstream 2.0.2, matching the upstream version packaged as `2.0.2-3` in Ubuntu Resolute
- preserve the remaining ament-specific namespace comment and standard literal namespace policies
- adapt the ament wrapper to cpplint 2.0.2's filename-aware `_ShouldPrintError()` API
- add regression coverage for the embedded version, C++20 headers, and lint error reporting

## Source and retained patches

The new `cpplint.py` comes from the official cpplint 2.0.2 tag:

[https://github.com/cpplint/cpplint/blob/2.0.2/cpplint.py](https://github.com/cpplint/cpplint/blob/2.0.2/cpplint.py)

Several historical ament patches are already present upstream, including `.hh` header recognition, automatic C-file handling, qualified `operator++` / `operator--` handling, and clang-analyzer `NOLINT` categories.

The update retains ament's remaining policy differences:

- require `// namespace ...` for named namespace closing comments
- allow standard literal and placeholder namespace using-directives in source files, while continuing to reject them in headers

cpplint 2.0 added filename-specific filter selectors, which changed `_ShouldPrintError()` from three arguments to four. The wrapper now forwards the filename so lint errors can be collected without raising `TypeError` and file-specific filters work correctly.

---

I have manually validated the update, and it works as expected. In particular, I no longer need to add `NOLINT` for the C++20 headers I tested.

As expected, though, updating `cpplint` also introduces some new or changed lint rules, which would require changes across a number of ROS 2 core repositories. I am opening this PR mainly to get some feedback on how we should approach those follow-up changes.

This also supersedes https://github.com/ament/ament_lint/pull/562

### DIFF
--- a/ament_cpplint/ament_cpplint/cpplint.py
+++ b/ament_cpplint/ament_cpplint/cpplint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 #
 # Copyright (c) 2009 Google Inc. All rights reserved.
 #
@@ -28,7 +28,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-# https://github.com/cpplint/cpplint/blob/6b1d29874dc5d7c3c9201b70e760b3eb9468a60d/cpplint.py
+# https://github.com/cpplint/cpplint/blob/2.0.2/cpplint.py
 
 """Does google-lint on c++ files.
 
@@ -43,7 +43,10 @@ We do a small hack, which is to ignore //'s with "'s after them on the
 same line, but it is far from perfect (in either direction).
 """
 
+from __future__ import annotations  # PEP 604 not in 3.9
+
 import codecs
+import collections
 import copy
 import getopt
 import glob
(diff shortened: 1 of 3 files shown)


mini-1235:
I think we would need to spend quite a bit of time fixing the new cpplint issues if we bump it to 2.0.2. Given that, does it make more sense to invest that effort in switching to more modern tooling, such as ament_clang_format + ament_clang_tidy, instead?

But we will need to update cpplint eventually anyway, since 2.0.2 is the version shipped by Ubuntu Resolute. 

Thoughts?

cc @ahcorde since I think you are working on the C++20 migration across the core repositories.

PR #570, opened 2026-01-30
xmllint: fetch external schemas via Python

mjcarroll:
Built-in support for fetching network-based schemas via the native http client in xmllint was removed in 2.15.0.

Ref: https://discourse.gnome.org/t/libxml2-2-15-0-released/31397

Built-with: Gemini Flash 3

PR #589, opened 2026-09-04
Report every file checked by ament_xmllint

mjcarroll:
`report.append((filename, errors))` and the `os.path.relpath` line above it sit outside the `for filename in files:` loop, so `report` only ever receives the last file that was checked and xmllint errors in every other file are silently discarded — the linter returns 0 for genuinely invalid XML. This was introduced by #570, which wrapped the loop in a `TemporaryDirectory` and re-indented the `for` statement without re-indenting the tail of its body; both lines move back inside the loop here.

This also fixes a second, smaller bug in the same function: `main()` is annotated `-> Literal[0, 1]` but returns a string when xmllint is not installed, which reaches callers as the confusing `assert "Could not find 'xmllint' executable" == 0`. It now prints to stderr and returns 1, matching the `'No files found'` branch above it.

### DIFF
--- a/ament_xmllint/ament_xmllint/main.py
+++ b/ament_xmllint/ament_xmllint/main.py
@@ -74,7 +74,8 @@ def main(argv: list[str] = sys.argv[1:]) -> Literal[0, 1]:
 
     xmllint_bin = shutil.which('xmllint')
     if not xmllint_bin:
-        return "Could not find 'xmllint' executable"
+        print("Could not find 'xmllint' executable", file=sys.stderr)
+        return 1
 
     report = []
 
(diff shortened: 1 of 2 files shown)


mjcarroll:
```
--packages-up-to ament_xmllint ament_cmake_xmllint pluginlib tracetools_launch
--packages-select ament_xmllint ament_cmake_xmllint pluginlib tracetools_launch
```

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30435/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23131/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10235/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29310/)

ahcorde (approved):

